PR #780, opened 2024-09-06
feat: Add native podman support

dpeckett:
Add native support for podman, tested/verified with podman on linux.


elliottneilclark:
Please give this a rebase, and collapse the NoisySocketImage into a single constant.

dpeckett:
I think there is some debate about whether this is strictly necessary at this point. I think in all environments where we would want to run a gateway we should have the docker compatible API available. And the extra code probably isn't all that worth it.

So I think I'll close this branch as an ultimately unnecessary experiment :).

PR #900, opened 2024-09-23
chore: release 0.18.1

elliottneilclark:
Summary:
Set version to 0.19.0 for next version
I'll push a tag for the commit right before this

Test Plan:
CI


elliottneilclark:
@JTarasovic I'm bumping the version since I changed the jwk keys. That will break stable versions API and we'll need to reset prod.

JTarasovic (approved):

Issue #425, opened 2024-07-17
Container command/args


elliottneilclark:
Currently we offer the ability for users to set the command and for each container. That input is not working because the inputs aren't hooked up. We need to have a better UI for this.

https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/

- Command and Args are both arrays (Though I can not think of a time when you need both to be an array). You can do something like:
```
command: ["/my/binary"]
args: ["\"${ENV_VALUE}\"", "test", "100"]
```

or
```
command: ["/my/binary", "\"${ENV_VALUE}\"", "test", "100"]
args: nil
```
- Phoenix forms do not accept a single input for an array field
- Most users won't want to set this

I think a reasonable middle ground for usability is to accept one path input. Use that as the first element in the command array. Then allow adding args in a sub-form manner.

PR #903, opened 2024-09-23
Support path and arguments for containers

jsonmaur:
Addresses #425 by replacing the `command` field in the container form with a virtual `path` field, and adds an input list for container arguments.

Screenshot 2024-09-23 at 16 42 16
Screenshot 2024-09-23 at 16 43 17

### DIFF
--- a/platform_umbrella/apps/common_ui/lib/common_ui/components/input_list.ex
+++ b/platform_umbrella/apps/common_ui/lib/common_ui/components/input_list.ex
@@ -107,28 +107,24 @@ defmodule CommonUI.Components.InputList do
     <div>
       <div :if={@label} class={label_class()}><%= @label %></div>
 
-      <%= if @field.value == [] do %>
-        <input type="hidden" name={"#{@field.name}[]"} />
-      <% else %>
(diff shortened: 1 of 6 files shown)


platform_umbrella/apps/common_ui/lib/common_ui/components/input_list.ex
```diff
@@ -107,28 +107,24 @@ defmodule CommonUI.Components.InputList do
...
-        <% end %>
-      <% end %>
+
+          <.button
+            variant="minimal"
+            icon={:x_mark}
+            phx-value-index={index}
+            phx-click={@remove_click}
+            phx-target={@phx_target}
+          />
```
elliottneilclark, line 125: If we use the drop param names this might be able to delete the last one?

elliottneilclark (approved):

platform_umbrella/apps/common_ui/lib/common_ui/components/input_list.ex
jsonmaur: See https://github.com/batteries-included/batteries-included/issues/904#issuecomment-2369790876

PR #905, opened 2024-09-23
feat: use postmark to deliver on prod

elliottneilclark:
Summary:
Set up swoosh to use the correct finch pool
Set up swoosh to use postmark for home base web prod
Set up a black hole server for home base

Test Plan:
Get a prod server up and running again.

### DIFF
--- a/platform_umbrella/apps/home_base_web/config/runtime.exs
+++ b/platform_umbrella/apps/home_base_web/config/runtime.exs
@@ -37,6 +37,11 @@ config :common_core, CommonCore.JWK,
   sign_key: :environment,
   verify_keys: [:home_a_pub, :home_b_pub]
 
+config :home_base, HomeBase.Mailer,
+  adapter: Swoosh.Adapters.Postmark,
+  # Fall back to the black hole postmark server if no key is provided
+  api_key: System.get_env("POSTMARK_KEY") || "d382f183-4b4b-417f-82db-3f9635a05c8b"
+
 config :home_base, HomeBase.Repo,
   database: postgres_database,
   hostname: postgres_host,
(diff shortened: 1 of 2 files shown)


elliottneilclark:
The key that's there is only for a black hole server that never sends out. In production we will need to set that environment variable to a real value. Similar to how we will do for jwt keys.

platform_umbrella/apps/home_base_web/config/runtime.exs
```diff
@@ -37,6 +37,11 @@ config :common_core, CommonCore.JWK,
   sign_key: :environment,
   verify_keys: [:home_a_pub, :home_b_pub]
 
+config :home_base, HomeBase.Mailer,
+  adapter: Swoosh.Adapters.Postmark,
+  # Fall back to the black hole postmark server if no key is provided
+  api_key: System.get_env("POSTMARK_KEY") || "d382f183-4b4b-417f-82db-3f9635a05c8b"
```
jsonmaur, line 43: Isn't the blackhole server meant to test bounces? Might be better to use `POSTMARK_API_TEST`, which returns valid API responses but doesn't actually send anything (https://postmarkapp.com/blog/handling-email-in-your-test-environment)

jsonmaur (approved):

platform_umbrella/apps/home_base_web/config/runtime.exs
elliottneilclark: This is a sandbox server that I created in postmark. It will return success in the Api, like their test env, the only difference is that we can see the sent emails in our account. I invited you and @JTarasovic to the postmark account.

Issue #663, opened 2024-08-20
Style error pages


elliottneilclark:
- [ ] Istio No Healthy Upstream page https://github.com/istio/istio/issues/10543#issuecomment-921179277
- [x] Contorl Server Phoenix 404
- [x] Control Server Phoenix 500
- [x] Home Base Phoenix 404
- [x] Home Base Phoenix 500

PR #913, opened 2024-09-25
Add custom 404 and 500 error pages

jsonmaur:
This PR addresses #663 by adding custom 404 and 500 error pages to Home Base and Control Server. Still need to implement the Istio No Healthy Upstream page, but waiting on further instructions. Since it sounds like that will be static HTML embedded into a yaml file, we should be able to copy the 500 page over and tweak the wording (https://batteries-included.slack.com/archives/C077UVAP13M/p1727224643573729).

---

### Control Server 404
Screenshot 2024-09-24 at 16 37 22

### Home Base 404
Screenshot 2024-09-24 at 18 00 02

### Control Server/Home Base 500
Screenshot 2024-09-24 at 16 36 23


elliottneilclark (approved):
The SVGs match the style. Thanks!

PR #915, opened 2024-09-25
feat: use the current request host, port, and scheme for starting

elliottneilclark:
Summary:
When coming to home base, we can have a lot of hostnames. So figure that
out at run time. However live view doesn't really offer that since the
socket persists. Instead we use the last on we saw before leaving
Plug.Conn goodbye.

Test Plan:
```
/bin/bash -c "$(curl -fsSL http://home.127-0-0-1.batrsinc.co:4100/installations/batt_01922808981a728a95959a2488018987/script)"
```


JTarasovic (approved):
I'm not sure this is going to work. I had issues with `request_url/1` returning http as istio is terminating TLS but I like the idea of determining it once and attaching it to the request.

elliottneilclark:
You're right I should use the x forwarded proto for the proto. Let me try and get that.

Issue #892, opened 2024-09-23
Batteries Home link on Control Server home page should point to a better url


elliottneilclark:
http://control.127-0-0-1.batrsinc.co:4000/magic -> http://home.127-0-0-1.batrsinc.co:4900/

That doesn't work, it should

PR #943, opened 2024-09-27
fix(magic): use correct home base URLs

JTarasovic:
fixes: #892

### DIFF
--- a/platform_umbrella/apps/common_core/lib/common_core/et/urls.ex
+++ b/platform_umbrella/apps/common_core/lib/common_core/et/urls.ex
@@ -2,7 +2,7 @@ defmodule CommonCore.ET.URLs do
   @moduledoc false
   alias CommonCore.Batteries.BatteryCoreConfig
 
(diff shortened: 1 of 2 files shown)


platform_umbrella/apps/common_core/lib/common_core/et/urls.ex
```diff
@@ -2,7 +2,7 @@ defmodule CommonCore.ET.URLs do
   @moduledoc false
   alias CommonCore.Batteries.BatteryCoreConfig
 
-  @local_home "http://home.traditional-service.127-0-0-1.batrsinc.co:4100/api/v1"
+  @local_home "http://home.127-0-0-1.batrsinc.co:4100/api/v1"
```
JTarasovic, line 5: Not sure what we want to do here...

JTarasovic (commented):
Tested locally and in AWS and the URLs look good.

jsonmaur (approved):

PR #965, opened 2024-10-03
chore(bi): upgrade to k8s 1.31

JTarasovic:
Test plan:
- [x] Verify cluster is created and bootstraps successfully

### DIFF
--- a/bi/pkg/cluster/kind/kind.go
+++ b/bi/pkg/cluster/kind/kind.go
@@ -16,7 +16,7 @@ import (
 )
 
 const (
-	KindImage         = "kindest/node:v1.30.0"
+	KindImage         = "kindest/node:v1.31.0"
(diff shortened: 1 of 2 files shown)


elliottneilclark (approved):
I assume there are no images that also need to move.

JTarasovic:
Good call. I'll try to spin up everything else to make sure but the install came up perfect so at least a good chunk are good.

elliottneilclark (changes requested):
Sorry for the whiplash. Please also update `pkg/cluster/kind/kind.go` to use the corresponding kind version for kubernetes and make sure that all comes up ok.

JTarasovic:
Lol. Already done. Pushing shortly.

elliottneilclark (approved):

PR #976, opened 2024-10-07
ci: test

JTarasovic:


elliottneilclark:
```
compile: writing output: write $WORK/b004/_pkg_.a: no space left on device
```

What is `$WORK` poinint to and how much is there when we fail? Maybe we're writing temporary .a's that get linked into a a place with less space than we need. Something is caching that data and if building for all the arches is too much? For some diffs we would be creating a larger set of .a's, too large to finish ?

Issue #957, opened 2024-10-03
For BI never give out more than a /24 to the ip address pools


elliottneilclark:
Even if there's a large range of IP's we should choose a constrained set. That allows a lot easier choice of which ones to use.

PR #1011, opened 2024-10-09
fix(bi): don't give out more than a /24

JTarasovic:
fixes: #957


JTarasovic:
Actually, I think I have a better way of doing this...

elliottneilclark (approved):

Issue #458, opened 2024-07-20
Make kube bootstrap wait for control server's pod to be stable for ~10 seconds


elliottneilclark:
Right now we only wait for the deployment to be healthy. However, the control server needs to start and then it will sometimes flap. We should wait for it to be healthy for a few seconds.

PR #1073, opened 2024-10-18
fix(bi): double check that bootstrap has completed

JTarasovic:
closes: #458 

This isn't my favorite but it should get the job done.


elliottneilclark:
Should we make bi re-bootstrap if it failed?

JTarasovic:
> Should we make bi re-bootstrap if it failed?

Maybe. It probably wouldn't work without deleting some resources first. We could certainly add that at some point.

Issue #23, opened 2024-05-10
Clean up form, input, and container components


jsonmaur:
There is some weird overlap between the Form, Input, and Container components that needs to be cleaned up. The Form component is defining spacing and layouts that should really be defined in Container. The Input component is defining a lot of stuff with spacing/labels/errors that should also be extracted into another component. Maybe something like `Fieldset` or `InputGroup`?

PR #1089, opened 2024-10-22
Use alert component for input errors

jsonmaur:
This updates the Input component to use the Alert component (with a new `minimal` type) for errors. Also gets rid of the buggy `phx-feedback` tailwind variant.

Related to https://github.com/batteries-included/batteries-included/issues/23


platform_umbrella/apps/common_ui/assets/tailwind.config.js
```diff
@@ -70,9 +70,6 @@ module.exports = {
     //
     //     <div class="phx-click-loading:animate-ping">
     //
-    plugin(({ addVariant }) =>
-      addVariant('phx-feedback', ['label:not(.phx-no-feedback) &'])
-    ),
     plugin(({ addVariant }) =>
       addVariant('phx-no-feedback', ['.phx-no-feedback&', '.phx-no-feedback &'])
```
elliottneilclark, line 74: Do we still need this if we aren't using the inverse?

platform_umbrella/apps/common_ui/lib/common_ui/components/alert.ex
```diff
@@ -86,29 +85,46 @@ defmodule CommonUI.Components.Alert do
...
   end
 
-  defp type_class("inline"), do: "px-3.5 py-2.5"
-
-  defp type_class("fixed") do
-    "fixed shadow-lg z-50 max-w-96 px-5 py-4 shadow-xl shadow-gray-darkest/10 z-50 hidden"
+  defp variant_class("info", type) do
+    [
+      "text-sky-500 dark:text-white",
+      type != "minimal" && "bg-sky-50 dark:bg-sky-500 border border-sky-100 dark:border-none",
```
elliottneilclark, line 91: This pattern of `bg-sky-50 dark:bg-sky-500 dark:border-none` being added if the type is not minimal can be extracted to repeat yourself less.

elliottneilclark (commented):
Do you have a screenshot of it? Something that show the difference would be great.

jsonmaur:
Visual difference is minimal.

## Before

Screenshot 2024-10-22 at 14 39 54

## After

Screenshot 2024-10-22 at 14 39 33

platform_umbrella/apps/common_ui/assets/tailwind.config.js
jsonmaur: Yes, we still need the `phx-no-feedback:*` classes. I was just using `phx-feedback:*` in some places to simplify the long classnames, but it turns out that doesn't work very well.

elliottneilclark (approved):

PR #1096, opened 2024-10-23
feat: add a healthz controller

elliottneilclark:
Summary:

- Checks that there are some kube state pods
- Checks that sql is up
- Add that as a liveness and readiness probe

Tests:
- `curl -vvv http://control.127-0-0-1.batrsinc.co:4000/healthz`
- Saw that fail with no pods when startin up
- Passes after 1 second
- Once this is in master and an image is built I'll bix start local


JTarasovic (approved):
I don't think that any of the specs changed? Not sure if we need to commit if not.

PR #1097, opened 2024-10-23
Use Field component in Postgres forms

jsonmaur:
- Replaces old Input component with Field component in Postgres forms
- Restructures the panels so instance count is in its own panel
- Removes `label_note` attribute that is no longer used

Related to https://github.com/batteries-included/batteries-included/issues/23

---

Screenshot 2024-10-23 at 10 56 23


elliottneilclark (approved):
The same visually but with the new components. Thanks

PR #1108, opened 2024-10-23
New Main Landing Page

YosmanProgrammer:


elliottneilclark (changes requested):
Under `One-click deployments with security, scale, and monitoring out of the box`
 That should be three columns but it's rendering as three rows.
 
 
image


Each of the images 
image


Should be much smaller to allow two column rendering like the document specified.

static/astro.config.mjs
```diff
@@ -57,15 +59,8 @@ export default defineConfig({
       })
     ),
     tasks(),
-    import.meta.env.PROD &&
```
elliottneilclark, line 60: What's causing the miss compile? Is the column selector in the css when you run npm run build ?

Playform compression does images, js, css, and html. Does this do that too?

What's the usage and track record of each?

Trading a dependency that we've used for a while needs a little why.

static/src/pages/index.astro
```diff
@@ -71,63 +75,26 @@ const metadata = {
...
       },
       {
-        title: 'Advanced Web Services',
+        title:
+          'Deploy web services at the speed of your users with autoscaling',
         description:
-          'Deploy from Docker to Knative with SSL, easier than ever. Get superior service on your own hardware thanks to our hands-off workflow.',
-        icon: 'checkoutline',
+          'Build and rollout your serverless apps with dynamic scaling and security. Choose private cloud or on-prem.',
+        icon: 'autoscaling',
```
elliottneilclark, line 91: Knative please

static/astro.config.mjs
YosmanProgrammer: Captura de pantalla 2024-10-24 a la(s) 2 42 37 a m
Fixed, put the compress on the last line

YosmanProgrammer: Sorry, I will be more careful with the changes I make, I will inform you first before making any changes.

elliottneilclark (approved):
This is an improvement. Thanks!

PR #1130, opened 2024-10-28
Home base fixes

jsonmaur:
- Adds redirect to new installation page on login if there are no installations
- Fixes confirm email, forgot password, and reset password pages to be in a panel
- Fixes typos in new installation descriptions and bug where it wasn't showing some sections
- Gets rid of "Welcome back" flash message when logging in
- Hides help icon when there is a flash message


elliottneilclark (changes requested):
CI clean please

elliottneilclark (approved):

PR #1157, opened 2024-11-04
feat(home): soft delete installations

JTarasovic:


JTarasovic:
Am able to delete an install with a report.

```
select r.id, r.report, i.slug, i.deleted_at from stored_usage_reports as r left join installations as i on r.installation_id = i.id;
─[ RECORD 1 ]────────────
id         │ 0192f95b-ba27-7250-a334-fb0507f2ef6c
report     │ {"batteries": ["istio", "istio_gateway", "battery_core", "metallb", "cloudnative_pg", "stale_resource_cleaner"], "node_report": {"avg_mem": 67368525824.0, "avg_cores": 16.0, "pod_counts": {"dev-control-plane": 15}}, "num_projects": 0, "redis_report": {"instance_counts": {}}, "ollama_report": {"model_counts": {}, "instance_counts": {}}, "knative_report": {"pod_counts": {}}, "postgres_report": {"instance_counts": {"internal.controlserver": 1}}, "namespace_report": {"pod_counts": {"battery-base": 3, "battery-core": 1, "battery-istio": 2}}, "traditional_services_report": {"instance_counts": {}}}
slug       │ dev
deleted_at │ 2024-11-04 22:49:50.338355
```

And it doesn't show in the UI
image

platform_umbrella/apps/home_base/test/home_base/customer_installs_test.exs
```diff
@@ -132,6 +133,10 @@ defmodule HomeBase.CustomerInstallsTest do
       assert_raise Ecto.NoResultsError, fn ->
         CustomerInstalls.get_installation!(installation.id)
       end
+
+      all_installations = Repo.all(Installation, with_deleted: true)
+      assert length(all_installations) == 1
+      assert installation.slug == all_installations |> List.first() |> Map.get(:slug)
```
elliottneilclark, line 139: Please add a unit test that tests this soft delete of an installation more explicitly.

platform_umbrella/apps/home_base/priv/repo/migrations/20241104220527_add_soft_delete.exs
```diff
@@ -0,0 +1,14 @@
+defmodule HomeBase.Repo.Migrations.AddSoftDelete do
+  use Ecto.Migration
+  import Ecto.SoftDelete.Migration
+
+  @existing_tables ~w(installations stored_host_reports stored_usage_reports teams teams_roles users)a
```
elliottneilclark, line 5: Only installations have all the soft delete columns in the schemas. We should bring all the tables up to finish. Stored usage reports and stored host reports need both a soft and time-based hard delete.

elliottneilclark (changes requested):
I like the soft deletes,  I think this is a little under ready. Not all the structs are wired up and there aren't tests showing that soft delete works for those structs. (create, read, delete, can't read, can read with `with_deleted: true`, etc)

platform_umbrella/apps/home_base/test/home_base/customer_installs_test.exs
JTarasovic: Testing that `deleted_at` is set? I'm not sure I'm tracking otherwise.

elliottneilclark: A test that inserts an installation, Then deletes it. Then tries to read to make sure that the soft deletes will work.
A test to make sure that we can read soft deletes. Insert an installation. delete it then read it with `with_deleted: true`.

You're wiring up soft deletes. We want to make sure that they don't get broken for any of the structs that use it.

platform_umbrella/apps/home_base/lib/home_base/accounts.ex
```diff
@@ -65,6 +65,10 @@ defmodule HomeBase.Accounts do
     Repo.all(User)
   end
 
+  def delete_user(%User{} = user) do
+    Repo.soft_delete(user)
+  end
```
elliottneilclark, line 70: Should we simply https://github.com/revelrylabs/ecto_soft_delete/tree/master?tab=readme-ov-file#repos

`use Ecto.SoftDelete.Repo` in `HomeBase.context/0`  or someplace like that?

elliottneilclark (approved):

elliottneilclark:
I assume that you tested this with deleting an install through the UI locally?

PR #1197, opened 2024-11-12
feat(bix): prevent running control server inadvertantly

JTarasovic:
I think we want a separate command for using control server against a remote cluster. Trying to get it working so that one could specify a slug was getting a bit messy and convoluted. The steps are largely the same but the postgres connection details would need to be figured out and passed. I think we can land this to keep us safe and add the remote cluster stuff later.


platform_umbrella/apps/kube_services/lib/kube_services.ex
```diff
@@ -2,16 +2,4 @@ defmodule KubeServices do
...
-
-  def get_current_events do
-    conn = CommonCore.ConnectionPool.get!()
-
-    operation = K8s.Client.list("v1", "Event")
-
-    with {:ok, result} <- K8s.Client.run(conn, operation) do
-      result
-    end
-  end
```
JTarasovic, line 14: Not used anywhere in our codebase.

platform_umbrella/apps/kube_services/lib/kube_services.ex
```diff
@@ -2,16 +2,4 @@ defmodule KubeServices do
...
-
-  def cluster_type, do: Application.get_env(:kube_services, :cluster_type, :dev)
```
JTarasovic, line 16: Was only used in `KubeServices.Connection` below.

platform_umbrella/apps/kube_services/lib/kube_services/connection.ex
```diff
@@ -1,12 +0,0 @@
-defmodule KubeServices.Connection do
```
JTarasovic, line 1: Not used anywhere.

elliottneilclark (approved):

elliottneilclark:
Thank you!

PR #1226, opened 2024-11-18
chore: regenerate openapi schema for keycloak

elliottneilclark:
Summary:
Keycloak now has an openapi schema. So this is the work needed to use
our existing code to regenerate with the latest

Test Plan:

Used it


JTarasovic (approved):

elliottneilclark:
They have some new Enums that I don't handle.

PR #1232, opened 2024-11-19
feat(sso): allow requiring MFA

JTarasovic:
Allows turning on MFA for `batterycore` realm.

Test plan:
- [x] enable keycloak - MFA not enabled
- [x] enable SSO, no MFA - MFA not enabled
- [x] enable SSO + MFA - MFA enabled 
- [x] enable SSO, go back and forth on MFA - MFA is enabled / disabled appropriately

### DIFF
--- a/platform_umbrella/apps/common_core/lib/common_core/batteries/sso_config.ex
+++ b/platform_umbrella/apps/common_core/lib/common_core/batteries/sso_config.ex
@@ -4,7 +4,7 @@ defmodule CommonCore.Batteries.SSOConfig do
   use CommonCore, :embedded_schema
 
   batt_polymorphic_schema type: :sso do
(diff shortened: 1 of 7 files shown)


platform_umbrella/apps/common_core/lib/common_core/batteries/sso_config.ex
```diff
@@ -4,7 +4,7 @@ defmodule CommonCore.Batteries.SSOConfig do
   use CommonCore, :embedded_schema
 
   batt_polymorphic_schema type: :sso do
-    defaultable_field :dev, :boolean, default: true
```
JTarasovic, line 7: I couldn't find anywhere where we were actually using this...

elliottneilclark (approved):

PR #1258, opened 2024-11-25
chore: update easy images

elliottneilclark:
Summary:
Update images for batteries that don't need crd or related updates

Test Plan:

- Started dev locally
- CI
- Started dev with redis, postgres, grafana, vm, etc

### DIFF
--- a/platform_umbrella/apps/common_core/lib/common_core/defaults/images.ex
+++ b/platform_umbrella/apps/common_core/lib/common_core/defaults/images.ex
@@ -5,7 +5,7 @@ defmodule CommonCore.Defaults.Images do
 
   @batteries_included_base "#{CommonCore.Version.version()}-#{CommonCore.Version.hash()}"
 
-  @cert_manager_allowed_tags ~w(v1.15.4)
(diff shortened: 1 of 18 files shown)


platform_umbrella/apps/common_core/lib/common_core/defaults/images.ex
```diff
@@ -5,8 +5,8 @@ defmodule CommonCore.Defaults.Images do
 
   @batteries_included_base "#{CommonCore.Version.version()}-#{CommonCore.Version.hash()}"
 
-  @cert_manager_allowed_tags ~w(v1.15.1)
-  @cert_manager_default_tag "v1.15.1"
+  @cert_manager_allowed_tags ~w(v1.15.4)
```
JTarasovic, line 8: Should we just be adding `.4` to the list of allowed tags instead of just replacing `.1`?

JTarasovic (approved):

platform_umbrella/apps/common_core/lib/common_core/defaults/images.ex
elliottneilclark: Nice catch. Thanks!

PR #1293, opened 2024-12-03
docs: use new youtube embed links for web and main flow

elliottneilclark:
Summary:
We have a couple of new videos to demo our flow. Since they are up we
should use them on the website. This PR updates the web and index pages
to use the new embed code

Test Plan:
- Tested locally
- netlify
- verify that deploy preview embeds the new vides for both pages

### DIFF
--- a/static/src/pages/index.astro
+++ b/static/src/pages/index.astro
@@ -29,7 +29,7 @@ const metadata = {
     <div class="mt-12 mx-auto max-w-screen-md">
       <div class="flex justify-center">
         <VideoModal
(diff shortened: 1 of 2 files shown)


JTarasovic (approved):

jfaz1:
Looks good! The only thing I noticed is the embedded video on the home page is the old one, not sure if that's intentional or not. 
image

elliottneilclark:
> Looks good! The only thing I noticed is the embedded video on the home page is the old one, not sure if that's intentional or not.
> 
> image

Nice catch. Updated.

jfaz1 (approved):

Issue #1176, opened 2024-11-11
monitoring: better tuning of storage defaults


JTarasovic:
We use 1GB and 5GB volumes for vm_{select,storage} respectively with a 1month (default) retention.

We should scale the storage and retention with the cluster size?

```
 -retentionPeriod value
     Data with timestamps outside the retentionPeriod is automatically deleted. The minimum retentionPeriod is 24h or 1d. See also -retentionFilter
     The following optional suffixes are supported: s (second), h (hour), d (day), w (week), y (year). If suffix isn't set, then the duration is counted in months (default 1)
```

elliottneilclark:
While you're doing this we should also make sure that the volume sizes are in integers when stored so that I can put a slider on it for the ui.

PR #1298, opened 2024-12-04
feat(monitoring): use preset sizes for victoria metrics

JTarasovic:
part of: #1176 

This doesn't use sliders yet. Just preset sizes.


JTarasovic:
Can use custom size with validation to prevent shrinking volumes:
image

Using preset:
image

elliottneilclark (approved):
This is awesome!

PR #1305, opened 2024-12-05
feat(monitoring): use sliders and presets for victoria metrics storage

JTarasovic:
closes: #1176


JTarasovic:
@elliottneilclark This is working except that the sliders lose their position so when you use the slider, it's correct.
image

But when you switch to the other slider or save and re-edit, they move back to the center position and I'm not sure why.
image

elliottneilclark (approved):
Nice. Thanks!

PR #1431, opened 2025-01-06
feat: use String.chunk to safely decode logs

elliottneilclark:
Summary:
- Only keep printable chunks of logs

Test Plan:

- CI


elliottneilclark:
```
K8s.Client.run(conn, get_operation(opts))
```

is typespeced to return reference or map. However it can return a binary string.

elliottneilclark:
This needs `coryodaniel/k8s#354` before it can go in cleanly

PR #1458, opened 2025-01-10
feat: add ability to restrict batteries by usage

JTarasovic:
This filters out batteries from the UI if the install's usage isn't allowed. This is a form of course grained feature flagging so that we can work on features without them being generally available.

Does not currently restrict battery from being installed just removes it from the UI. We could always add that if needed.
Are there any other places where we should filter / hide these batteries?

### DIFF
--- a/platform_umbrella/apps/control_server_web/lib/control_server_web/components/projects/batteries_form.ex
+++ b/platform_umbrella/apps/control_server_web/lib/control_server_web/components/projects/batteries_form.ex
@@ -15,12 +15,14 @@ defmodule ControlServerWeb.Projects.BatteriesForm do
 
   def mount(socket) do
     groups = Catalog.groups_for_projects()
+    core_battery = KubeServices.SystemState.SummaryBatteries.core_battery()
 
     {:ok,
      socket
      |> assign(:class, nil)
      |> assign(:tab, :required)
-     |> assign(:groups, groups)}
+     |> assign(:groups, groups)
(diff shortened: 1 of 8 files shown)


JTarasovic:
SSO battery not installable when not allowed
image

Promtail and SSO not available though project flow when not allowed:
image
image

platform_umbrella/apps/control_server_web/lib/control_server_web/components/projects/batteries_form.ex
```diff
@@ -15,12 +15,16 @@ defmodule ControlServerWeb.Projects.BatteriesForm do
 
   def mount(socket) do
     groups = Catalog.groups_for_projects()
+    core_battery = KubeServices.SystemState.SummaryBatteries.core_battery()
+    # core_battery / config / usage will only be nil during some tests :grimacing:
```
elliottneilclark, line 19: How is that possible? It seems so fundamental. Can you add how that happens as a comment until it gets cleaned up.

elliottneilclark (approved):

platform_umbrella/apps/control_server_web/lib/control_server_web/components/projects/batteries_form.ex
JTarasovic: It was a nightmare but I was able to fix the tests so that we don't this anymore.

PR #1485, opened 2025-01-19
Add initial Forgejo docs

jfaz1:


elliottneilclark:
Lets hold off on landing this one. The docs are good, but the battery needs some work. We should mark it as development only right now. @JTarasovic just landed the ability to do that for a battery.

jfaz1:
> Lets hold off on landing this one. The docs are good, but the battery needs some work. We should mark it as development only right now. @JTarasovic just landed the ability to do that for a battery.

Gotcha, I'll just leave it in draft mode till it makes sense to publish!

elliottneilclark (approved):

PR #1496, opened 2025-01-21
chore: mark forgejo battery for internal development only at this time

JTarasovic:


elliottneilclark (approved):

elliottneilclark:
LGTM Thanks. We can change this when we correctly route ssh port.

PR #1508, opened 2025-01-22
feat(project_export): migrate from modal to a separate export page

JTarasovic:


JTarasovic:
Adds the export button to the project table(s) (if enabled):
image


A new export page is used instead of a modal. This sets the groundwork to display the fields so that the user can select which to export. It currently just dumps out JSON.
image

If one ended up on the page some other way, we display a message that the feature isn't enabled
image

elliottneilclark (approved):

PR #1509, opened 2025-01-22
chore: upate jupyter notebook image

elliottneilclark:
Summary:
Update to the new register

Test Plan:
- Started one

### DIFF
--- a/platform_umbrella/apps/common_core/lib/common_core/notebooks/jupyter_lab_notebook.ex
+++ b/platform_umbrella/apps/common_core/lib/common_core/notebooks/jupyter_lab_notebook.ex
@@ -68,7 +68,13 @@ defmodule CommonCore.Notebooks.JupyterLabNotebook do
 
   batt_schema "jupyter_lab_notebooks" do
     slug_field :name
-    field :image, :string, default: "quay.io/jupyter/datascience-notebook:lab-4.3.4"
+
+    field :image, :string,
+      default:
+        :jupyter_datascience_lab
+        |> CommonCore.Defaults.Images.get_image!()
+        |> CommonCore.Defaults.Image.default_image()
+
     field :storage_size, :integer
     field :storage_class, :string
     field :cpu_requested, :integer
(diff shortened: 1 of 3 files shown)


platform_umbrella/apps/common_core/lib/common_core/notebooks/jupyter_lab_notebook.ex
```diff
@@ -68,7 +68,7 @@ defmodule CommonCore.Notebooks.JupyterLabNotebook do
 
   batt_schema "jupyter_lab_notebooks" do
     slug_field :name
-    field :image, :string, default: "jupyter/datascience-notebook:lab-4.0.7"
+    field :image, :string, default: "quay.io/jupyter/datascience-notebook:lab-4.3.4"
```
JTarasovic, line 71: Should this go in the image "registry"?

elliottneilclark: Yes, let me get that up.

platform_umbrella/apps/common_core/lib/common_core/notebooks/jupyter_lab_notebook.ex
```diff
@@ -68,7 +68,13 @@ defmodule CommonCore.Notebooks.JupyterLabNotebook do
 
   batt_schema "jupyter_lab_notebooks" do
     slug_field :name
-    field :image, :string, default: "jupyter/datascience-notebook:lab-4.0.7"
+
+    field :image, :string,
```
JTarasovic, line 72: `defaultable_image_field` should do this for you if you wanna do it that way...

JTarasovic (approved):

platform_umbrella/apps/common_core/lib/common_core/notebooks/jupyter_lab_notebook.ex
elliottneilclark: I don't. For jupyter notebooks they should be immutable. Scientific users expect that their experiments don't change because a platform upgrades. Additionally I don't want to restrict the image to only being this one, there are a lot of jupyter notebook docker images and users can choose theirs.

My ideal would be that we have a `image_field` type that also has the existing form, while allowing an override of the full image path.

PR #1532, opened 2025-01-28
chore(karpenter): upgrade to 1.2.0

JTarasovic:
Upgrade karpenter.


platform_umbrella/apps/common_core/lib/common_core/resources/karpenter.ex
```diff
@@ -399,8 +401,9 @@ defmodule CommonCore.Resources.Karpenter do
     cluster_name = Core.config_field(state, :cluster_name)
 
     spec = %{
-      "amiFamily" => "AL2",
       "role" => battery.config.node_role_name,
+      "amiFamily" => "AL2",
+      "amiSelectorTerms" => [%{"alias" => "al2@v20250123"}],
```
JTarasovic, line 406: This will need to change / we'll need to figure out a way to update the version that should be used. We could use `@latest` but that has its own set of issues.

elliottneilclark: Should this be a defaultable field on the config, then?

JTarasovic: The concern with that is that it would need to be changed to the correct version by the user as they upgrade, I think.

elliottneilclark: If they don't set the field, they will get the default of the running version. If they do set the `ami_selector` (or whatever) field then they are responsible for making sure it works. That seems reasonable since that ami seems to be tied to the running karpenter version

elliottneilclark:
With the change in ami from previous version to now, will upgrades be smooth?

elliottneilclark (approved):

JTarasovic:
I think so but lemme double check.

elliottneilclark:
> I think so but lemme double check.

PWHEW

JTarasovic:
Tested against a cluster running 1.0.8 and it updates to 1.2.0 with no issues. :relieved: 
Also tested a new cluster w/ 1.2.0 and it still spins up fine.

elliottneilclark (approved):

elliottneilclark:
public.ecr.aws/karpenter/controller:1.0.8 is running in prod.

PR #1541, opened 2025-01-30
chore(karpenter): upgrade to 1.0.8

JTarasovic:
This works on an existing cluster to upgrade it as well as for a brand new cluster.


bi/pkg/cluster/pulumi.go
```diff
@@ -81,7 +81,7 @@ func (p *pulumiProvider) configure(ctx context.Context) (auto.Workspace, error)
 		"cluster:maxSize":        {Value: "4"},
 		"cluster:minSize":        {Value: "2"},
 		"cluster:name":           {Value: p.slug},
-		"cluster:version":        {Value: "1.32"},
+		"cluster:version":        {Value: "1.31"},
```
elliottneilclark, line 84: We'll need to update our cluster manually for now.

elliottneilclark (approved):

Issue #1543, opened 2025-01-30
Shouldn't push latest oci tag on release


elliottneilclark:
When the github action build actions run, we shouldn't push the `latest` tag since there's already a more recent build.

PR #1544, opened 2025-01-30
fix: don't push latest tag on tag releases

elliottneilclark:
Summary:
When releasing a version we push that. However there's already a latest
tag somewhere else. So we shouldn't make that tag regress.

This should fix: #1543

Test Plan:
-  CI ?

### DIFF
--- a/.github/workflows/bix_docker.yml
+++ b/.github/workflows/bix_docker.yml
@@ -72,7 +72,7 @@ jobs:
           echo "PUSH=true" >> "$GITHUB_ENV"
 
       # set ADDITIONAL_TAGS to either latest or the current tag
-      - if: ${{ fromJSON(env.PUSH) }}
+      - if: ${{ fromJSON(env.PUSH) && !startsWith(github.ref, 'refs/tags/') }}
         run: |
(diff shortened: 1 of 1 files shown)


JTarasovic:
`latest` doesn't get pushed on release. `ADDITIONAL_TAGS` gets "overwritten" in the next step. 

```
  env:
    PUSH: true
    ASDF_DIR: /home/runner/.asdf
    ASDF_DATA_DIR: /home/runner/.asdf
    ACTIONS_RUNNER_ACTION_ARCHIVE_CACHE: /opt/actionarchivecache
    ACTIONS_RUNTIME_URL: ...
    ACTIONS_RUNTIME_TOKEN: ***
    ACTIONS_CACHE_URL: ...
    ACTIONS_ID_TOKEN_REQUEST_URL: ...
    ACTIONS_ID_TOKEN_REQUEST_TOKEN: ***
    ACTIONS_RESULTS_URL: ...
    ADDITIONAL_TAGS: 0.50.0
```

JTarasovic:
But it doesn't hurt to not write it, I suppose.

JTarasovic (approved):

elliottneilclark:
Ohhhh. I missed that it was creating a file that will be sourced.

Issue #1510, opened 2025-01-22
Configure Jupyter with overrides.json


elliottneilclark:
- Turn off shut down menu item
- Turn off logout
- No news

PR #1629, opened 2025-02-14
feat(notebooks): configure jupyter with overrides

JTarasovic:
closes: #1510


JTarasovic:
There is no notification about news or checking for updates when one first logs in and the "log out" and "shutdown" options are removed from the file menu.

image

elliottneilclark (approved):

PR #1635, opened 2025-02-17
ci: upgrade asdf and pin version

JTarasovic:
This gets rid of the warning spam about upgrade to 0.16.x. The actions don't currently support it so pin 1 version back and we can upgrade again once the actions are ready.


JTarasovic:
Request was approved and a gpu node was created

image

elliottneilclark (approved):
Thank you. That was annoying

PR #1638, opened 2025-02-18
chore(karpenter): enable gpu support for nvidia nodes

JTarasovic:
The AL2 AMIs don't have default support for GPUs but the bottlerocket AMIs do. So create a nodeclass for it and use that class for the nvidia node pool.


elliottneilclark (approved):
GPU's are getting so close

PR #1646, opened 2025-02-19
fix(bootstrap): use aws provided csi policy

JTarasovic:
The EBS CSI driver controller updated and needs new permissions. Instead of maintaining the policy ourselves, let Amazon handle updating the policy.


elliottneilclark (approved):
gogogo

PR #1647, opened 2025-02-20
fix(karpenter): match instance metadata options to bootstrap values

JTarasovic:
I noticed that the aws-load-balancer-controller was having issues retrieving data from the metadata service when it was running on karpenter nodes. They initially run on bootstrap nodes which allow the containers to retrieve metadata.

### DIFF
--- a/platform_umbrella/apps/common_core/lib/common_core/resources/karpenter_pools.ex
+++ b/platform_umbrella/apps/common_core/lib/common_core/resources/karpenter_pools.ex
@@ -151,6 +151,9 @@ defmodule CommonCore.Resources.KarpenterPools do
       "role" => role,
       "securityGroupSelectorTerms" => [%{"tags" => %{"karpenter.sh/discovery" => cluster_name}}],
       "subnetSelectorTerms" => [%{"tags" => %{"karpenter.sh/discovery" => cluster_name}}],
+      # allow containers to access the AWS metadata service
+      # necessary for e.g. aws-load-balancer-controller to run on karpenter nodes
+      # this also matches how we configure the bootstrap nodes
       "metadataOptions" => %{"httpPutResponseHopLimit" => 2},
       "tags" => %{
         "karpenter.sh/discovery" => cluster_name,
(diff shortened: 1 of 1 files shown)


platform_umbrella/apps/common_core/lib/common_core/resources/karpenter_pools.ex
```diff
@@ -161,4 +143,21 @@ defmodule CommonCore.Resources.KarpenterPools do
...
   end
+
+  defp build_nodeclass_spec(family, alias, role, cluster_name) do
+    %{
+      "amiFamily" => family,
+      "amiSelectorTerms" => [%{"alias" => alias}],
+      "role" => role,
+      "securityGroupSelectorTerms" => [%{"tags" => %{"karpenter.sh/discovery" => cluster_name}}],
+      "subnetSelectorTerms" => [%{"tags" => %{"karpenter.sh/discovery" => cluster_name}}],
+      "metadataOptions" => %{"httpPutResponseHopLimit" => 2},
```
JTarasovic, line 154: I refactored this spec generation but setting this hop limit to 2 is the fix.

JTarasovic:
I can scale up the deployment and it will schedule across all nodes and we can see that it is running.

```
k get po -n battery-base -l app=aws-load-balancer-controller -o custom-columns="NAME:.metadata.name,NODE:.spec.nodeName,STATUS:.status.phase"
NAME                                          NODE                                           STATUS
aws-load-balancer-controller-875b64dd-25g8h   ip-100-64-101-239.us-east-2.compute.internal   Running
aws-load-balancer-controller-875b64dd-t6dpk   ip-100-64-101-95.us-east-2.compute.internal    Running
aws-load-balancer-controller-875b64dd-vwltx   ip-100-64-103-174.us-east-2.compute.internal   Running
aws-load-balancer-controller-875b64dd-wjrwk   ip-100-64-102-88.us-east-2.compute.internal    Running
```

platform_umbrella/apps/common_core/lib/common_core/resources/karpenter_pools.ex
elliottneilclark: Please comment on why this is needed and how it fixes it.

elliottneilclark (approved):

PR #1649, opened 2025-02-20
feat(ml): allow notebooks to use GPU nodes

JTarasovic:
Requires using a GPU enabled image e.g. `quay.io/jupyter/tensorflow-notebook:cuda-latest`


JTarasovic:
:raised_hands: 

image

platform_umbrella/apps/common_core/lib/common_core/notebooks/jupyter_lab_notebook.ex
```diff
@@ -82,6 +82,8 @@ defmodule CommonCore.Notebooks.JupyterLabNotebook do
     field :memory_requested, :integer
     field :memory_limits, :integer
 
+    field :node_type, Ecto.Enum, values: ~w(default nvidia_gpu)a, default: :default
```
JTarasovic, line 85: How do we like this for the field? How do we want to expose this in the UI? Just a dropdown?

elliottneilclark: SHould they choose which GPU they want here? In that case then yeah it will be a drop down. If it's just on or off then I would make this boolean and use a switch.

JTarasovic:
This is how the form looks:

image

elliottneilclark (approved):

PR #1662, opened 2025-02-26
fix: put nvidia device plugin behind flag

JTarasovic:
The bottlerocket instances that we're using for GPUs already run the device plugin so we don't need to run that separately. 

There's also some cleanup for feature discovery as I was trying to get the device plugin working separately from the underlying instance.


elliottneilclark (approved):
Yeah that battery might need to be removed if we don't use it.

PR #1685, opened 2025-03-04
Add BI overview post

jfaz1:
This PR adds a general overview of the BI architecture and how it works.

I made a couple of educated guesses with the graphs, please let me know if you would like anything changed :smile:


static/src/content/doc/how-it-works.md
```diff
@@ -0,0 +1,420 @@
...
+- **Kubernetes Updates**: For managed installations, the underlying Kubernetes
+  cluster receives automatic updates to maintain security and stability.
+- **Routine Maintenance**: Batteries Included handles tasks like log rotation,
+  certificate renewal, and database optimization automatically.
+
+The combination of proactive monitoring and seamless maintenance creates a
+self-sustaining environment where potential problems are identified early and
+routine updates occur without disruption. By automating these traditionally
+time-consuming aspects of infrastructure management, teams can focus on
+application development rather than operational overhead.
```
elliottneilclark, line 382: This is different from the other sections. It's about the batteries

jfaz1:
Thanks for the detailed feedback! I went ahead and applied it, let me know if you want any other changes made, and if the diagrams look good now.

static/src/content/doc/how-it-works.md
jfaz1: Sorry, how do you mean? Do you want me to merge this section with the Battery one or just remove it?

elliottneilclark: Probably remove I think. Is there a compelling reason to call out metrics that hard in a how it works article?

elliottneilclark:
> I was thinking Development

That was my first thought as well; lets go with development.

static/src/content/doc/how-it-works.md
jfaz1: Yeah I think you're right, just thought maybe someone reading it might be wondering how the metrics-side works but I think it's clear from other call-outs in the post.

elliottneilclark (approved):

PR #1696, opened 2025-03-06
fix(ollama): don't allow conflicting GPU choices

JTarasovic:


JTarasovic:
Now we get UI alerts if we try to do something funny:
image
image

elliottneilclark (approved):

PR #1698, opened 2025-03-07
feat(ollama): smarter preset selection based on models

JTarasovic:
If you choose a model that's bigger than the preset, we'll update to the smallest preset that makes sense for the model. If you've chosen a preset that's bigger than the model, we keep it.


JTarasovic:
smarter-model-presets

elliottneilclark (approved):

PR #1712, opened 2025-03-11
feat: don't snapshot password_versions

elliottneilclark:
Summary:
- Don't take a snapshot of password versions field in postgres
- Don't insert password_version from a snapshot

Test Plan:
Two unit tests added
- Test one shows there's no password_version in the snapshot
- Test two shows that the password_version isn't used in importing the
  snapshot


platform_umbrella/apps/control_server/test/control_server/projects/project_snapshot_test.exs
```diff
@@ -36,7 +36,7 @@ defmodule ControlServer.Projects.ProjectSnapshotTest do
 
   defp small_pg_project do
     project = insert(:project)
-    pg_cluster = insert(:postgres_cluster, project_id: project.id, virtual_size: "small")
+    pg_cluster = :postgres_cluster |> insert(project_id: project.id, virtual_size: "small") |> dbg()
```
JTarasovic, line 39: Not sure if this should be committed.

JTarasovic (approved):

PR #1714, opened 2025-03-11
fix(bi): prevent wireguard log spam on stop

JTarasovic:


elliottneilclark (approved):
Thank you, clean up needed some love!

PR #1715, opened 2025-03-12
chore: upgrade local tool versions

JTarasovic:


elliottneilclark (approved):
Oh nice, we can update Erlang because we changed the curve we use for elliptic keys!

PR #1719, opened 2025-03-12
feat(postgres): display backups in UI

JTarasovic:
Adds a table with the backups for a cluster to the cluster show page. This will be used for recovery in a future change.

### DIFF
--- a/platform_umbrella/apps/control_server_web/lib/control_server_web/live/postgres/show.ex
+++ b/platform_umbrella/apps/control_server_web/lib/control_server_web/live/postgres/show.ex
@@ -40,9 +40,9 @@ defmodule ControlServerWeb.Live.PostgresShow do
      |> assign_k8_pods()
      |> assign_timeline_installed()
      |> assign_grafana_url()
+     |> assign_backups()
      |> maybe_assign_edit_versions()
-     |> maybe_assign_events()
-     |> maybe_assign_backups()}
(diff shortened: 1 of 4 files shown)


JTarasovic:
Shown on the "show" page:

With backups:
image

Without backups:
image

platform_umbrella/apps/control_server_web/lib/control_server_web/live/postgres/show.ex
```diff
@@ -229,6 +241,7 @@ defmodule ControlServerWeb.Live.PostgresShow do
         >
           Edit Versions
         </:tab>
+        <:tab selected={@live_action == :backups} patch={backups_url(@cluster)}>Backups</:tab>
```
elliottneilclark, line 244: Can we hide this link if there are no backups available

elliottneilclark (approved):

platform_umbrella/apps/control_server_web/lib/control_server_web/live/postgres/show.ex
JTarasovic: Not sure the juice was worth the squeeze but it's hidden now:
https://github.com/batteries-included/batteries-included/pull/1719/commits/92a4a0936cbdc00eeefe706fd752d5ce3422e0b1

PR #1743, opened 2025-03-20
refactor: add ability to get polymorphic mappings at runtime

JTarasovic:


elliottneilclark:
What's needing this?

JTarasovic:
> What's needing this?

I'm working on setting up cluster backup configuration and, looking at the system_battery setup, there's going to be some duplication in processing the mappings. I was thinking / hoping that we'd be able to consolidate / generate the functions at some point in the future. This would just be groundwork for that.

elliottneilclark (approved):

PR #1744, opened 2025-03-20
feat(bix): add ability to run test setup independently

JTarasovic:
We used to have this with nix but it got dropped in the conversion but I find myself running `ex-test-deep` fairly regularly just to set up the DB.

### DIFF
--- a/bin/bix
+++ b/bin/bix
@@ -88,6 +88,7 @@ Available commands:
 
 **Elixir Commands**
 - ex-test-deep      Run all tests with coverage and reset the database
+- ex-test-setup     Setup the test database
 - ex-test           Run all tests
 - ex-lint           Run all linters (dialexir, credo, format)
 - ex-credo          Run credo linter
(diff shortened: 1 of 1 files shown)


bin/bix
```diff
@@ -561,13 +562,24 @@ do_ex_test_deep() {
 
     export MIX_ENV=test
     run_mix "do" deps.get, compile --warnings-as-errors >/dev/null
```
elliottneilclark, line 564: Is it possible to use the setup function here

elliottneilclark (approved):

bin/bix
JTarasovic: You mean like have a function that just does the mix commands to set up the database? And both `ex-test-deep` and `ex-test-setup` would call that?

Or something else?

elliottneilclark: I don't think we even need a new function. If deep test calls the new function with the right args I think we can remove some duplicate code

JTarasovic: I think I'm tracking. Lemme know if this is along the lines of what you're thinkin'.

elliottneilclark (approved):

PR #1749, opened 2025-03-24
feat(cnpg): allow minimal backup configuration

JTarasovic:
Allows configuring whether backups should be enabled or not. Does so in a way that is extensible as we expose additional backup configuration.

### DIFF
--- a/platform_umbrella/apps/common_core/lib/common_core/postgres/backup_config.ex
+++ b/platform_umbrella/apps/common_core/lib/common_core/postgres/backup_config.ex
@@ -1,4 +1,4 @@
-defmodule CommonCore.Postgres.BackupConfig do
+defmodule CommonCore.Postgres.PGBackupConfig do
   @moduledoc false
   use CommonCore, :embedded_schema
 
--- a/platform_umbrella/apps/common_core/lib/common_core/postgres/backup_config.ex
+++ b/platform_umbrella/apps/common_core/lib/common_core/postgres/backup_config.ex
@@ -1,0 +1,17 @@
+defmodule CommonCore.Postgres.BackupConfig do
+  @moduledoc false
+  use CommonCore, :embedded_schema
+
+  import CommonCore.Util.Tuple
+
+  @backup_types [
+    none: "None",
+    object_store: "Object Store"
+  ]
+
+  batt_embedded_schema do
+    field :type, Ecto.Enum, values: Keyword.keys(@backup_types)
+  end
+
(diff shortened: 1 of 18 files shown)


platform_umbrella/apps/common_core/lib/common_core/postgres/cluster.ex
```diff
@@ -121,6 +126,16 @@ defmodule CommonCore.Postgres.Cluster do
     end)
   end
 
+  defp maybe_put_backup_config(changeset) do
+    case get_field(changeset, :backup_config) do
+      nil ->
+        put_embed(changeset, :backup_config, %BackupConfig{type: :none})
```
elliottneilclark, line 132: I don't think we want this; it will mean anytime there's a changeset on an existing pg cluster we always have to pass in the current backup config. Meaning there will be no way to do a partial edit.

It's probably better to add this onto the reading side.

platform_umbrella/apps/common_core/lib/common_core/postgres/backup_config.ex
```diff
@@ -0,0 +1,17 @@
+defmodule CommonCore.Postgres.BackupConfig do
```
elliottneilclark, line 1: The other embedded structs are prefixed with PG should we do that here to follow, or should we re-name others since they are already in the Postgres namespace?

platform_umbrella/apps/common_core/lib/common_core/postgres/cluster.ex
JTarasovic: Yeah, we may not even need to. If we keep the semantic that nil is equivalent to none, right?

platform_umbrella/apps/common_core/lib/common_core/resources/postgres/cloudnative_pg_clusters.ex
```diff
@@ -151,6 +153,7 @@ defmodule CommonCore.Resources.CloudnativePGClusters do
     |> B.spec(spec)
     |> F.require_non_nil(battery.config.bucket_name)
     |> F.require_non_nil(battery.config.service_role_arn)
+    |> F.require(cluster.backup_config.type == :object_store)
```
elliottneilclark, line 156: Remember that backup_config might be nil.
```suggestion
    |> F.require(cluster.backup_config != nil && cluster.backup_config.type == :object_store)
```

elliottneilclark (approved):

Issue #1762, opened 2025-03-25
seed data getting deleted


JTarasovic:
Not sure what's going on here but I'm guessing it's a bit of a race

JTarasovic:
Control server was cleaning up unmanaged resources. It just took long enough (because of the DB connection problem) that we noticed it.

PR #1773, opened 2025-03-26
fix(int-prod): prevent seed data from being deleted too soon

JTarasovic:
Fixes: #1762


JTarasovic:
I'm not sure that I'm tracking. The logic is in `KubeServices.Stale`. If it hasn't been enough time, the resource isn't listed as potentially stale...

elliottneilclark:
I was thinking we say if it has this tag and the creation timestamp + the diff hasn't passed then don't list this resource as stale yet in `stale?`.

Something like `KubeServices.Stale.tagged_and_time_remaning?`. That way, the stale page won't list this resource as a resource to be deleted until it's ready to be cleaned up.

elliottneilclark (approved):

PR #1775, opened 2025-03-28
fix(bi): update signature algorithm to match new home-base

JTarasovic:
The algorithm changed but we didn't update it on the `bi` side which is preventing installations from installing.


elliottneilclark (approved):
WOW Nice catch!

PR #1776, opened 2025-03-28
chore: bump bi stable version to 0.58.0

elliottneilclark:
Summary:
I missed the version and it isn't signing well. So fix that.

Test Plan:
- /shrug
- CI


JTarasovic:
We only have binaries for 0.57.0. Do we need to do a release to go along with this as well?

elliottneilclark:
Yep we can land this as the PR right before the release.

JTarasovic (approved):

PR #1779, opened 2025-03-28
fix(control): don't mark resources stale during snapshot

JTarasovic:


platform_umbrella/apps/kube_services/lib/kube_services/stale/stale.ex
```diff
@@ -26,6 +28,7 @@ defmodule KubeServices.Stale do
 
   def stale?(resource, seen_res_set \\ nil)
   def stale?(resource, nil), do: stale?(resource, recent_resource_map_set())
+  def stale?(resource, seen_res_set) when seen_res_set == @empty, do: stale?(resource, recent_resource_map_set())
```
JTarasovic, line 31: Is there a concern here of infinite regression in case the recent resources is persistently empty? Seems unlikely...

JTarasovic:
```elixir
iex(8)> KubeServices.Stale.stale?(r, MapSet.new())
true
iex(9)> recompile
==> kube_services
Compiling 1 file (.ex)
Generated kube_services app
:ok
iex(10)> KubeServices.Stale.stale?(r, MapSet.new())
false
iex(11)> KubeServices.Stale.stale?(r, MapSet.new([]))
false
```

platform_umbrella/apps/kube_services/lib/kube_services/stale/stale.ex
elliottneilclark: I think you're right, and it's safe.

elliottneilclark (approved):
Go Go.

PR #1781, opened 2025-03-28
Add single-node post

jfaz1:
This PR adds a single-node BI page explaining local development workflow and how it scales. I made a note to revisit this once we have configuration migration working (though that should also have its own page anyways).

### DIFF
--- a/static/src/content/doc/single-node.md
+++ b/static/src/content/doc/single-node.md
@@ -29,8 +29,8 @@ but also ensuring what works locally works in production as well.
 ## Getting Started Locally
 
 Unlike traditional approaches that require developers to set up a bunch of
-dependencies or write complex Docker Compose files or Kubernetes manifests,
-Batteries Included provides a streamlined setup process:
+dependencies or write complex Docker Compose/Kubernetes manifests, Batteries
+Included provides a streamlined setup process:
 
 1. Visit [home.batteriesincl.com](https://home.batteriesincl.com) and create a
    new installation.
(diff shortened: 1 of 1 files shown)


static/src/content/doc/single-node.md
```diff
@@ -0,0 +1,134 @@
...
+journey:
+
+1. **Local Development**: Start with a lightweight environment locally.
+2. **Staging, Testing**: Deploy to staging with the same configuration.
+3. **Production Deployment**: Confidently launch to production using the same
+   familiar tools.
+
+Using Batteries Included like this eliminates the common frustration of "works
+on my machine" problems, not only reducing onboarding time for new team members
+but also ensuring what works locally works in production as well.
```
elliottneilclark, line 27: For now we should drop staging/testing. Then use this as a template for the article.

Local Development should focus on starting, port-forwarding, and deploying batteries. Debugging and other related topics should also be covered. Let's talk about how to get the logs rather than just hinting that we have them. 
Production should focus on how you use the same tools to add more production features. Batteries like monitoring, keycloak, etc., are not likely to be used on local dev but critical for prod.

static/src/content/doc/single-node.md
```diff
@@ -0,0 +1,134 @@
...
+WireGuard VPN configurations using:
+
+```bash
+bi vpn config my-production-installation
+```
+
+Batteries Included provides this consistent, easy-to-use toolset across all
+stages of development, making the transition from local to production as smooth
+as possible, and ensuring that what works locally will work in production just
+the same.
```
elliottneilclark, line 134: This isn't something that users will be doing much at all. They don't need to know this to see that Batteries Included scales up. If we want to talk about the VPN then we should include that as a separate doc.

elliottneilclark (changes requested):

PR #1791, opened 2025-03-31
fix(bootstrap): only bootstrap what's needed for controlserver

### LINKED ISSUE #1790: Bootstrap startup secret issues
I noticed that starting a secured production install with keycloak didn't successfully run keycloak.

We bootstrap the secret with a password and keycloak starts. 
That password is never persisted anywhere so when we actually seed from the install summary a different password IS stored to the DB but Keycloak is using the original from bootstrap.

For security, we wouldn't want to include the password in the install spec.

JTarasovic:
Closes #1790

Test plan:
- [x] bootstrap locally
- [x] run install from home-base but `--skip-bootstrap` and finish against this PR.

### DIFF
--- a/platform_umbrella/apps/common_core/lib/common_core/resources/sso/keycloak.ex
+++ b/platform_umbrella/apps/common_core/lib/common_core/resources/sso/keycloak.ex
@@ -65,6 +65,8 @@ defmodule CommonCore.Resources.Keycloak do
     |> B.name("keycloak")
     |> B.namespace(namespace)
     |> B.data(data)
+    |> F.require_non_nil(battery.config.admin_password)
+    |> F.require(battery.config.admin_password != "")
   end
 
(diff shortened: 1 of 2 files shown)


platform_umbrella/apps/common_core/lib/common_core/resources/sso/keycloak.ex
```diff
@@ -65,6 +65,7 @@ defmodule CommonCore.Resources.Keycloak do
     |> B.name("keycloak")
     |> B.namespace(namespace)
     |> B.data(data)
+    |> F.require(battery.config.admin_password != "")
```
JTarasovic, line 69: Belt and suspenders. Just prevents secret from being created ( and keycloak from starting) if there's not an admin password.

elliottneilclark (approved):

PR #1810, opened 2025-04-02
ci: make reusable actions for set up

JTarasovic:


.github/workflows/cache.yml
```diff
@@ -62,37 +62,18 @@ jobs:
...
         run: bin/bix ex-dialyzer
 
       - name: Run tests for caching
-        run: bin/bix ex-test-deep
+        shell: bash
+        env:
+          MIX_ENV: test
+        run: |
+          pushd platform_umbrella
+          mix compile --force
```
elliottneilclark, line 76:
```suggestion
          pushd platform_umbrella
          mix "do", deps.get, compile --force
```

elliottneilclark: If the deps change then the cache won't have the deps there so we always need to ask to get the deps before compiling.

JTarasovic: Yeah, I was thinking the dialyzer step beforehand would do it but it wouldn't pull test-only dependencies. :facepalm:

elliottneilclark (approved):

PR #1845, opened 2025-04-08
fix: keep battery_core battery starting first no matter what

elliottneilclark:
Summary:
- This should fix oauth enabled clusters

Test Plan:
- tried the repro locally and it passed


JTarasovic (approved):
Works locally for me as well.

PR #1848, opened 2025-04-08
feat: enable postgres restore from backup

JTarasovic:


platform_umbrella/apps/control_server_web/lib/control_server_web/live/postgres/show.ex
```diff
@@ -34,7 +34,6 @@ defmodule ControlServerWeb.Live.PostgresShow do
     {:noreply,
      socket
      |> assign(:restore_form, nil)
-     |> assign(:restore_enabled, SummaryBatteries.battery_installed(:postgres_restore))
```
elliottneilclark, line 37: Kind clusters can't backup or restore. Is there some way for us not to show things that aren't possible for the user?

elliottneilclark (approved):

platform_umbrella/apps/control_server_web/lib/control_server_web/live/postgres/show.ex
JTarasovic: It doesn't show if there aren't any backups. And we don't allow enabling backups on a cluster unless it's running in AWS.

JTarasovic:
Double checked against a cluster with backups and everything works as expected.

PR #1851, opened 2025-04-09
chore: refactor json logging configuration

JTarasovic:
Move the formatter configuration to runtime config. Moving to `new/1` should ensure that we always have a valid formatter config. I think the issue we had was that `redactors` was `nil` instead of `[]` which I fixed (temporarily) by copying the output of new to the compile time config.

### DIFF
--- a/platform_umbrella/config/runtime.exs
+++ b/platform_umbrella/config/runtime.exs
@@ -3,3 +3,7 @@ import Config
 config :logger, :console,
   format: "$time $metadata[$level] $message\n",
   level: :debug
+
+if config_env() == :prod do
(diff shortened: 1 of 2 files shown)


platform_umbrella/config/runtime.exs
```diff
@@ -3,3 +3,7 @@ import Config
 config :logger, :console,
   format: "$time $metadata[$level] $message\n",
   level: :debug
+
+if config_env() == :prod do
+  config :logger, :default_handler, formatter: LoggerJSON.Formatters.Basic.new(metadata: [:mfa, :request_id])
```
JTarasovic, line 8: We can't call the function in config at compile time. It does work at runtime though.

elliottneilclark (approved):

PR #1858, opened 2025-04-11
ci: enable integration tests

JTarasovic:
Enable

### DIFF
--- a/bin/bix
+++ b/bin/bix
@@ -87,6 +87,7 @@ Available commands:
 - go-clean-build    Clean the temporary BI build directory
 
 **Elixir Commands**
+- ex-int-test       Run integration tests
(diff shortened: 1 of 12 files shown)


bin/bix
```diff
@@ -561,6 +562,13 @@ tag_push() {
     done
 }
 
+do_ex_int_test() {
+    export MIX_ENV=test
+    run_mix "do" deps.get, compile --warnings-as-errors >/dev/null
+    run_mix test --trace --warnings-as-errors --all-warnings --only cluster_test
```
elliottneilclark, line 568: From what I remember, we might need to run the test in apps/verify to ensure that other applications aren't started.

elliottneilclark (approved):

PR #1869, opened 2025-04-14
ci: invalidate elixir cache on bi changes

JTarasovic:
We cache the locally built `bi` binary in the elixir cache. When there's any change in that directory, the binary will get rebuilt which takes > 1 min in CI.

So use the hash of the `bi` directory as an input to the elixir cache key.

### DIFF
--- a/.github/actions/setup-asdf/action.yml
+++ b/.github/actions/setup-asdf/action.yml
@@ -13,6 +13,10 @@ runs:
       with:
         asdf_branch: ${{ inputs.asdf_branch }}
 
+    - name: Determine hash keys
+      shell: bash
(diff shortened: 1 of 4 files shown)


elliottneilclark (approved):
How do we end up with bi in the elixir cache? I would have expected that it's just the mix directory.

JTarasovic:
> How do we end up with bi in the elixir cache? I would have expected that it's just the mix directory.

https://github.com/batteries-included/batteries-included/actions/runs/14452239980/job/40527419457#step:5:14

It's actually the verify kind install worker that builds the `bi` binary during elixir tests. I'll set it up so that it doesn't do that.

PR #1878, opened 2025-04-16
fix(bi): better heuristic for control readiness

JTarasovic:


bi/pkg/specs/wait.go
```diff
@@ -44,7 +50,21 @@ func (spec *InstallSpec) WaitForBootstrap(ctx context.Context, kubeClient kube.K
...
+		slog.Error("Failed to get access info config map", slog.Any("err", err))
+		return err
+	}
+
+	// try to make sure the control server is up before we return
+	slog.Debug("Attempting to connect to control server", slog.String("url", info.GetURL()))
+	err = retry.Do(func() error {
+		_, err := http.Head(info.GetURL())
+		return err
+	}, retry.Context(ctx))
```
JTarasovic, line 65: This makes sure that we can actually connect to the control server after we've checked that the deployment is "finished" rolling out.

bi/pkg/specs/wait.go
```diff
@@ -93,36 +101,35 @@ func jobFailed(conditions []batchv1.JobCondition) bool {
...
+}
+
+func deployReady(deploy *appsv1.Deployment) (bool, error) {
+	status := deploy.Status
+
+	return status.ObservedGeneration > 1 &&
+			status.AvailableReplicas > 0 &&
+			status.UpdatedReplicas > 0 &&
+			status.ReadyReplicas > 0,
+		nil
```
JTarasovic, line 132: Change to watch the controlserver deployment instead of the pods. The `ObservedGeneration > 1` ensures that cs has started at least once and updated the deploy with the new secret key base, etc. The rest of the conditions make sure we have at least one ready pod on the newest version of the deploy.

bi/pkg/specs/wait.go
```diff
@@ -138,25 +145,30 @@ func batteryInfoConfigMapWatchOpts(ns string) *kube.WatchOptions {
...
+		if err != nil {
+			slog.Debug(
+				"failed to convert unstructured object into typed resource",
+				slog.String("namespace", u.GetNamespace()),
+				slog.String("name", u.GetName()),
+				slog.Any("err", err),
+			)
+			return false, nil
+		}
+		return fn(obj)
```
JTarasovic, line 172: This same pattern was repeated for each of the watches. Refactored and made it generic to keep things D.R.Y.

bi/pkg/specs/wait.go
elliottneilclark: Please add that info as a comment somewhere

elliottneilclark (approved):

PR #1901, opened 2025-04-21
chore: prepare for 0.67.0

elliottneilclark:
Summary:
- bix set-version
- This will land after the bi bump and the images are updated.

Test Plan:
- CI
- I'll cut a release assuming the verify is clean


elliottneilclark:
This will only land after the pg version bump and the bi version bump.

JTarasovic (approved):

PR #1919, opened 2025-04-24
fix(bi): don't create random spec.json directories

JTarasovic:


elliottneilclark (commented):
YAY. I had been meaning to track this down. Thank you!

elliottneilclark (approved):

PR #1921, opened 2025-04-24
ci(redis): add integration tests

JTarasovic:


platform_umbrella/apps/common_core/lib/common_core/installs/batteries.ex
```diff
@@ -12,7 +12,15 @@ defmodule CommonCore.Installs.Batteries do
   @aws_only_battery_types ~w(aws_load_balancer_controller karpenter)a
   @aws_battery_types ~w(battery_ca)a ++ @aws_only_battery_types
 
-  @internal_int_test_battery_types ~w(battery_core cloudnative_pg ferretdb istio_gateway metallb traditional_services)a
+  @internal_int_test_battery_types ~w(
```
elliottneilclark, line 15: Should we make the tests install the batteries? That way test installs don't have to run so many operators?

platform_umbrella/apps/verify/test/redis_test.exs
```diff
@@ -0,0 +1,90 @@
...
+    |> fill_in(@name_field, with: instance_name)
+    |> click(@save_button)
+    # verify show page
+    |> assert_has(show_page_header(instance_name))
+    # Assert that the first pod for the cluster is shown
+    |> assert_has(Query.css("tr:first-child", text: "#{instance_name}-0"))
+
+    # Assert that we have gotten to the show page
+    path = current_path(session)
+    assert path =~ ~r/\/redis\/[\d\w-]+\/show$/
```
elliottneilclark, line 31: We'll need some way to verify pods are started. That along with battery install will likely common patterns that we'll need

elliottneilclark (approved):

platform_umbrella/apps/common_core/lib/common_core/installs/batteries.ex
JTarasovic: Yes. I've been mulling it a bit. Going to poke around and see if I can figure out something reasonably ergonomic.

platform_umbrella/apps/verify/test/redis_test.exs
JTarasovic: Yeah, are you thinking that we'd use the k8s client to do that? Or go to the pod show page and check the status?

PR #1934, opened 2025-04-29
feat(verify): add ability to install batteries

JTarasovic:
- adds BatteryInstallWorker to install batteries necessary for a test module
- removes unnecessary batteries from bootstrapping for `int-test`
- refactors test helpers


elliottneilclark (approved):
YAY, this looks awesome

PR #1942, opened 2025-04-30
ci(int-test): ensure the correct kubeconfig

JTarasovic:


elliottneilclark (approved):
Looks good to me. Were we getting the wrong config sometimes?

JTarasovic:
@elliottneilclark When running locally, one could get the wrong config. For example, as I was running int-tests and bootstrapping a dev cluster, it got messed up.

I was also thinking we'd need the config for verifying the state of the cluster but we talked about using the pods page so that's not (currently) necessary.

PR #1943, opened 2025-04-30
chore: set up actionlint config

JTarasovic:
I have actionlint setup us as an automatic linter. This adds a basic config so that it works well here.


JTarasovic:
If we don't want to add this, would it be alright if I added it to `.gitignore` so that I could keep it locally?

elliottneilclark (approved):
If you think it's valuable then we should hook it up to run automatically.
https://github.com/rhysd/actionlint/blob/v1.7.7/docs/usage.md#use-actionlint-on-github-actions

PR #1945, opened 2025-05-01
fix(bi): don't install unneeded plugins

JTarasovic:
pulumi will install the needed plugins without us needing to keep them in sync.

Address some other `TODO`s while I was in there.

### DIFF
--- a/bi/pkg/cluster/pulumi.go
+++ b/bi/pkg/cluster/pulumi.go
@@ -8,6 +8,7 @@ import (
 
 	"bi/pkg/cluster/eks"
 	"bi/pkg/cluster/util"
+	"bi/pkg/specs"
 
 	"github.com/adrg/xdg"
 	"github.com/pulumi/pulumi/sdk/v3/go/auto"
@@ -23,10 +24,10 @@ const (
 
 type pulumiProvider struct {
 	initSuccessful bool
+	spec           *specs.InstallSpec
 
 	cfg auto.ConfigMap
 	projectName,
-	slug,
 	workDirRoot string
 
 	pulumiHome auto.LocalWorkspaceOption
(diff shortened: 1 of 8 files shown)


bi/pkg/cluster/pulumi.go
```diff
@@ -23,54 +24,49 @@ const (
...
-	// add plugins that need to be installed here
-	plugins := map[string]string{
-		"aws":       "v6.54.2",
-		"cloudinit": "v1.4.7",
-		"tls":       "v5.0.7",
-	}
-	if err := p.installPlugins(ctx, ws, plugins); err != nil {
-		return fmt.Errorf("failed to install necessary pulumi plugins: %w", err)
-	}
-
```
JTarasovic, line 59: These weren't kept in sync with the versions in `go.mod` so we were installing these versions and pulumi was installing the version matching `go.mod` anyways.

bi/pkg/cluster/pulumi.go
```diff
@@ -23,54 +24,49 @@ const (
...
-	slug,
 	workDirRoot string
 
 	pulumiHome auto.LocalWorkspaceOption
 	pulumi     auto.LocalWorkspaceOption
 	envVars    auto.LocalWorkspaceOption
 }
 
-func NewPulumiProvider(slug string) Provider {
+func NewPulumiProvider(spec *specs.InstallSpec) Provider {
```
JTarasovic, line 38: Passing the full spec allows us to start infering information from the spec instead of hardcoding.

bi/pkg/specs/battery_spec.go
```diff
@@ -25,10 +25,19 @@ func (s *InstallSpec) GetBatteryConfigField(typ, field string) (any, error) {
...
+
 func (s *InstallSpec) GetCoreNamespace() (string, error) {
 	ns, err := s.GetBatteryConfigField("battery_core", "core_namespace")
 	if err != nil {
 		return "", err
 	}
 	return ns.(string), nil
 }
+
+func (s *InstallSpec) GetCoreUsage() (string, error) {
```
JTarasovic, line 37: The usage is pulled a half dozen times or so already to wrap it up in a func for reuse.

elliottneilclark (approved):

PR #1964, opened 2025-05-05
refactor(bix): migrate to git style subcommands

JTarasovic:


bin/bi-source
```diff
@@ -0,0 +1,332 @@
...
+set -Eeuo pipefail
+
+# Get where the current script is located
+SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd -P)
+# Get the root directory of the project
+ROOT_DIR=$(cd "$SCRIPT_DIR/.." &>/dev/null && pwd -P)
+
+source "${ROOT_DIR}/bin/lib/common-functions.sh"
+
+format_mix_code() {
```
JTarasovic, line 12: I think some of these should probably be delegated to their respective binaries and "wrapped" here but that can be later.

elliottneilclark (changes requested):
Please change the README.md and `static/content/our_tools.md` since bootstrap and dev have changed.

Issue #1796, opened 2025-04-01
ferretdb


JTarasovic:
1. install battery
1. create PG cluster
1. create ferret instance
1. verify ferret pods created
1. check logs for anything obvious

PR #1976, opened 2025-05-07
feat(verify): add initial ferretdb tests

JTarasovic:
closes: #1796


elliottneilclark (approved):
Do we need to up the timeout now?

JTarasovic:
@elliottneilclark I think we'll be okay for this round. I'll update it with the next set of tests that I add.

PR #1984, opened 2025-05-08
fix(bix): exit gracefully on cleanup

JTarasovic:
I was focused on fixing the dangling port forward and missed that we're now killing ourselves and leaving a non-zero exit code.

```bash
zsh: terminated  bix m deps.get

 /batteries-included
❌143 ❯ bix m deps.get
```

### DIFF
--- a/bin/lib/common-functions.sh
+++ b/bin/lib/common-functions.sh
@@ -92,10 +92,11 @@ try_portforward() {
 }
 
 cleanup() {
+    local code="$?"
     trap - ERR EXIT
     # We're going to be sending SIGTERM to ourselves
     # Handle it gracefully
-    trap 'exit 0' SIGINT SIGTERM
+    trap 'exit "$code"' SIGINT SIGTERM
 
     # We can end up in cleanup from a few different places and colors might not be set
     safe_colors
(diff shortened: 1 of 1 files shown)


bin/lib/common-functions.sh
```diff
@@ -93,6 +93,9 @@ try_portforward() {
 
 cleanup() {
     trap - ERR EXIT
+    # We're going to be sending SIGTERM to ourselves
+    # Handle it gracefully
+    trap 'exit 0' SIGINT SIGTERM
```
elliottneilclark, line 98: Won't this hide the test exit code? We start the test with exit non-zero, but then trap runs, and we exit 0.

JTarasovic: Shit. Probably. Lemme fix.

JTarasovic:
```bash
 /batteries-included
❯ bix m bad
** (Mix) The task "bad" could not be found
[2025-05-08T15:31:23-0500]: Cleaning up all subprocesses and jobs

 /batteries-included
❌1 ❯ bix m compile
[2025-05-08T15:31:44-0500]: Cleaning up all subprocesses and jobs

 /batteries-included
❯ echo "$?"
0
```

elliottneilclark (approved):

PR #1986, opened 2025-05-08
feat: determine images from shared registry

JTarasovic:


platform_umbrella/apps/common_core/lib/common_core/defaults/images.ex
```diff
@@ -363,4 +365,17 @@ defmodule CommonCore.Defaults.Images do
     ver = batteries_included_version()
     "ghcr.io/batteries-included/home-base:#{ver}"
   end
+
+  def __after_compile__(_env, _bytecode) do
+    path = Path.relative_to_cwd("priv/images.json")
```
JTarasovic, line 370: This may not be the best (or even a good?) place or name. Happy to change if there are any better ideas.

elliottneilclark: Do we ever need this again? Seems like a one-shot thing. I would takethe nice json that you generated, load the file into a list via `IncludeResource`. For now we can hand edit the JSON.

JTarasovic: Are you thinking that we'd maintain the image "registry" in the JSON file and load it into Elixir as opposed to generating JSON from Elixir?

image_registry.yaml
```diff
@@ -0,0 +1,294 @@
...
+    - 1.2.0
+keycloak:
+  default_tag: 26.2.0
+  name: quay.io/keycloak/keycloak
+  tags:
+    - 25.0.2
+    - 25.0.6
+    - 26.2.0
+# Be sure to check:
+# https://kiali.io/docs/installation/installation-guide/prerequisites/#version-compatibility
```
JTarasovic, line 136: Copied. Wouldn't be possible in standard JSON.

image_registry.yaml
```diff
@@ -0,0 +1,294 @@
...
+  default_tag: *knative_istio_default_tag
+  name: gcr.io/knative-releases/knative.dev/net-istio/cmd/webhook
+  tags: *knative_istio_allowed_tags
+knative_serving_activator:
+  default_tag: &knative_default_tag v1.17.0
+  name: gcr.io/knative-releases/knative.dev/serving/cmd/activator
+  tags: &knative_allowed_tags
+    - v1.15.1
+    - v1.15.2
+    - v1.17.0
```
JTarasovic, line 166: Here's an example of the anchor and alias setup that aims to replicate how we did it with `@registry`

platform_umbrella/apps/common_core/lib/common_core/defaults/images.ex
```diff
@@ -3,316 +3,13 @@ defmodule CommonCore.Defaults.Images do
...
-      })
-  }
+  @batteries_included_base "#{CommonCore.Version.version()}-#{CommonCore.Version.hash()}"
 
   @spec get_image(atom()) :: Image.t() | nil
   def get_image(name) do
-    @registry
-    |> Enum.filter(fn {k, _v} -> k == name end)
-    |> Enum.map(&elem(&1, 1))
-    |> List.first()
```
JTarasovic, line 315: I'm not sure why this used a pipeline. I think it just didn't evolve with the registry? Or am I missing something?

platform_umbrella/apps/common_core/lib/common_core/defaults/registry.ex
```diff
@@ -0,0 +1,17 @@
+defmodule CommonCore.Defaults.Registry do
+  @moduledoc false
+
+  alias CommonCore.Defaults.Image
+
+  defmacro __before_compile__(_env) do
```
elliottneilclark, line 6: Why this rather than include resource?

elliottneilclark (approved):

platform_umbrella/apps/common_core/lib/common_core/defaults/registry.ex
JTarasovic: Yeah, here's a quick diff of that.

It ends up re-encoding the registry for every image (which happens about 50 times during compilation). /shrug

If you have a preference, we can go with that. I'm not too sussed either way.

```
diff --git a/platform_umbrella/apps/common_core/lib/common_core/defaults/images.ex b/platform_umbrella/apps/common_core/lib/common_core/defaults/images.ex
index 7604c48aa..32ad78db8 100644
--- a/platform_umbrella/apps/common_core/lib/common_core/defaults/images.ex
+++ b/platform_umbrella/apps/common_core/lib/common_core/defaults/images.ex
@@ -1,15 +1,19 @@
 defmodule CommonCore.Defaults.Images do
   @moduledoc false
 
-  alias CommonCore.Defaults.Image
+  use CommonCore.IncludeResource, registry: "../../../image_registry.yaml"
 
-  @before_compile CommonCore.Defaults.Registry
+  alias CommonCore.Defaults.Image
 
   @batteries_included_base "#{CommonCore.Version.version()}-#{CommonCore.Version.hash()}"
 
   @spec get_image(atom()) :: Image.t() | nil
   def get_image(name) do
-    Map.get(registry(), name, nil)
+    :registry
+    |> get_resource()
+    |> YamlElixir.read_from_string!()
+    |> Map.get(Atom.to_string(name))
+    |> Image.new!()
   end
 
   @spec get_image!(atom()) :: Image.t()
diff --git a/platform_umbrella/apps/common_core/lib/common_core/defaults/registry.ex b/platform_umbrella/apps/common_core/lib/common_core/defaults/registry.ex
deleted file mode 100644
index 09a4a78bb..000000000
--- a/platform_umbrella/apps/common_core/lib/common_core/defaults/registry.ex
+++ /dev/null
@@ -1,17 +0,0 @@
-defmodule CommonCore.Defaults.Registry do
-  @moduledoc false
-
-  alias CommonCore.Defaults.Image
-
-  defmacro __before_compile__(_env) do
-    registry =
-      "../../../image_registry.yaml"
-      |> Path.relative_to_cwd()
-      |> YamlElixir.read_from_file!()
-      |> Map.new(fn {key, val} -> {String.to_atom(key), Image.new!(val)} end)
-
-    quote do
-      def registry, do: unquote(Macro.escape(registry))
-    end
-  end
-end
```

elliottneilclark: This is fine, I was just wondering.

PR #2024, opened 2025-05-21
fix: line up kind, kind node, and k8s versions

JTarasovic:


bi/pkg/cluster/kind/kind.go
```diff
@@ -16,7 +16,9 @@ import (
 )
 
 const (
-	KindImage         = "kindest/node:v1.33.1"
+	// This ideally will be a pre-built image for the kind release we're using
+	// But should match the minor version of k8s that we're running in AWS
+	KindImage         = "kindest/node:v1.32.5"
```
JTarasovic, line 21: We're still using 1.32 in AWS. 1.33 isn't available yet AFAIK.

elliottneilclark (approved):

PR #2032, opened 2025-05-28
style: update .credo.exs

elliottneilclark:
Description:
- Update the .credo.exs
- Update a lot of the found nits
- This is mostly to re-enable dbg testing

Test Plan:
- CI still passes

### DIFF
--- a/platform_umbrella/config/dev.exs
+++ b/platform_umbrella/config/dev.exs
@@ -95,7 +95,8 @@ config :kube_services, cluster_type: :dev
 config :logger, :console,
   format: "$time $metadata[$level] $message\n",
   level: :debug,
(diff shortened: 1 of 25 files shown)


platform_umbrella/config/dev.exs
```diff
@@ -95,7 +95,8 @@ config :kube_services, cluster_type: :dev
 config :logger, :console,
   format: "$time $metadata[$level] $message\n",
   level: :debug,
-  metadata: [:mfa, :request_id],
+  metadata:
+    ~w(mfa request_id name namespace fail_cnt cluster retries_left type reason error errors realm action kind result pid id status pool)a,
```
JTarasovic, line 99: Do we want to add these to the json_logger config?

PR #2037, opened 2025-05-28
chore(bi): allow sideloading images into kind cluster

JTarasovic:


bi/pkg/log/log.go
```diff
@@ -25,7 +24,7 @@ func SetupLogging(verbosity string, color bool) error {
 		Level:      Level,
 		AddSource:  Level == slog.LevelDebug,
 		NoColor:    !color,
-		TimeFormat: time.Kitchen,
+		TimeFormat: "15:04:05.000",
```
JTarasovic, line 27: This matches the elixir log output format locally.

.tool-versions
```diff
@@ -11,3 +11,4 @@ python 3.12.9
 shfmt 3.11.0
```
JTarasovic, line 11: I don't see anywhere where we're actually using python. Do you remember what we're using it for?

elliottneilclark (approved):

.tool-versions
elliottneilclark: I think it was being used for one of the elixir libraries for building a NIF.

PR #2051, opened 2025-06-02
chore(verify): fix miscellaneous nits and try to get int-tests running faster

JTarasovic:
Passing the env through does seem to do something?

https://github.com/batteries-included/batteries-included/actions/runs/15398673928


elliottneilclark (approved):
Looks good to me. Since we are having to pass the env through so explicitly, it feels like we should make `image_tar_path` into a `bi start` flag at some point.

PR #2065, opened 2025-06-03
fix(bix): export version for int tests

JTarasovic:


JTarasovic:
It's not needed for the images. They are getting built and loaded. It's needed for `Verify`. The spec that it creates at runtime doesn't have the version we want unless we set VERSION_OVERRIDE.

PR #2067, opened 2025-06-03
chore(bix): miscellaneous nits and fixes

JTarasovic:


bin/bi-local
```diff
@@ -122,18 +122,18 @@ do_dev() {
 
 do_int_test() {
     local tag
-    tag=$(bix d get-image-tag)
+    tag=$(bi-docker get-image-tag)
```
JTarasovic, line 125: If we use `bix`, we'll do an extra arg parse and case traversal. Skip it and call the desired binary directly.

bin/bi-local
```diff
@@ -229,7 +229,7 @@ phx-server)
     do_phx_server
     ;;
 int-test)
-    do_int_test
+    do_int_test "${args[@]}"
```
JTarasovic, line 232: Allow passing args to the underlying `bix ex int-test` command to allow testing single modules or whatever.

bin/bi-registry
```diff
@@ -48,10 +48,13 @@ do_save_int_images() {
...
     log "Found ${#imgs[@]} images to save"
 
+    # docker removes registry portion from dockerhub hosted images
+    # so we won't find it and will try to re-pull below if we don't remove this
+    local dockerio='docker.io/'
     local pull_count
     pull_count=0
     for img in "${imgs[@]}"; do
-        if ! docker images "${img}" --format=json | jq -e '.' &>/dev/null; then
+        if ! docker images "${img/$dockerio/}" --format=json | jq -e '.' &>/dev/null; then
```
JTarasovic, line 57: Previously, it wouldn't "find" the istio images that are hosted on dockerhub even if they were already pulled. It's largely a cosmetic fix if the images are already there but it was confusing to me to see it in the logs.

elliottneilclark (approved):

PR #2068, opened 2025-06-03
ci: run int tests on PRs

JTarasovic:


.github/workflows/cache.yml
```diff
@@ -209,3 +222,47 @@ jobs:
...
+        run: |
+          bin/bix docker ensure-base-images
+          bin/bix docker build-image control-server
+          bin/bix docker build-image kube-bootstrap
+          bin/bix docker retag-images-for-ghcr
+
+      - name: Create image archive
+        run:
+          bin/bix registry save-int-images
+          /home/runner/work/_temp/base_images.tar
```
elliottneilclark, line 261: This includes the control-server and kube-bootstrap images, which means this shared cache will have a lot of overwrites. Should we add a `bin/bix registry save-base-int-images` command that includes the base image and the registry images?

Bascially I am worried this cache is going to leak data and become a write bottleneck.

bin/bi-docker
```diff
@@ -165,6 +165,17 @@ do_push_images() {
...
+    for image_name in kube-bootstrap control-server home-base pastebin; do
+        for tag in $(docker images "${image_name}" --format=json | jq -r '.Tag'); do
+            log "${GREEN}Tagging image ${BLUE}ghcr.io/batteries-included/${image_name}:${tag}${NOFORMAT}"
+            docker tag "${image_name}:${tag}" "ghcr.io/batteries-included/${image_name}:${tag}"
+        done
+    done
+
+    log "${GREEN}Finished retagging images${NOFORMAT}"
+}
+
```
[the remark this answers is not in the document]
JTarasovic: Yeah, I like that. There's a bunch of places where we have to re-tag and no places where we really use the bare image names.

.github/workflows/cache.yml
JTarasovic: Well, what I've currently got will only update if the registry changes or the dockerfiles change (basically we've updated the base images). It'd help to already have those layers present as it's less to pull and extract from the registry.

I'm not opposed to having a separate command for just the base images though. I had definitely thought about it.

PR #2081, opened 2025-06-06
chore(verify): cleanup knative sleeps

JTarasovic:


platform_umbrella/apps/verify/test/knative_test.exs
```diff
@@ -9,18 +9,12 @@ defmodule Verify.KnativeTest do
...
+    # wait a sec for knative to "install"
+    # it's actually quicker to pause here then to wait for future syncs
+    |> sleep(5_000)
     # trigger a new summary
     |> trigger_k8s_deploy()
-    # wait for a sec for it to complete
-    |> then(fn session ->
-      Process.sleep(1_000)
-      session
-    end)
```
JTarasovic, line 23: Since we're handling 404s in `assert_pods_in_deployment_running`, we don't need to wait here.

PR #2084, opened 2025-06-06
test(verify): disable dev shm usage in wallaby

elliottneilclark:
Description:
GitHub actions have a small /dev/shm ( I expect even the larger runner sizes). So this makes Chrome run through Chromedriver use /tmp

Test Plan:
- CI


JTarasovic (approved):
:crossed_fingers:

PR #2088, opened 2025-06-09
ci: run int tests on select PRs

JTarasovic:


.github/actions/restore-images/action.yml
```diff
@@ -0,0 +1,50 @@
+---
```
JTarasovic, line 1: New action to restore images since we're doing it a few times now.

PR #2097, opened 2025-06-09
ci: fix broken image push

JTarasovic:


elliottneilclark (approved):

JTarasovic:
We need to check out the repo so that the local action is accessible. _le sigh_

PR #2112, opened 2025-06-11
ci: fix artifact id passing

JTarasovic:
I have no idea what the problem was but this fixes it.


.github/workflows/bix_docker.yml
```diff
@@ -92,7 +92,7 @@ jobs:
         if: ${{ fromJSON(env.needs-archive) }}
         env:
           new-artifact-id: ${{ steps.upload-archive.outputs.artifact-id }}
-        run: echo "artifact-id=${new-artifact-id}" >> "$GITHUB_ENV"
```
JTarasovic, line 95: I don't understand why `${new-artifact-id}` wouldn't work but it outputs 'artifact-id'.
/shrug

PR #2118, opened 2025-06-11
fix: yaml parsing errors in dependabot

elliottneilclark:
Description:
Dependabot was complaining about our new time config: 
https://github.com/batteries-included/batteries-included/runs/43917758949

```
The property '#/updates/0/schedule/time' of type integer did not match the following type: string
The property '#/updates/1/schedule/time' of type integer did not match the following type: string
The property '#/updates/2/schedule/time' of type integer did not match the following type: string
The property '#/updates/3/schedule/time' of type integer did not match the following type: string
The property '#/updates/4/schedule/time' of type integer did not match the following type: string
The property '#/updates/5/schedule/time' of type integer did not match the following type: string
The property '#/updates/6/schedule/time' of type integer did not match the following type: string
The property '#/updates/7/schedule/time' of type integer did not match the following type: string
The property '#/updates/8/schedule/time' of type integer did not match the following type: string
```

That's because the yaml specs says that if the first character of a value is a digit then it's a number (float if there are .'s and integer otherwise). So this wraps the values, and staggers start time of anything that had a int-tet label

Test Plan:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,7 +5,7 @@ updates:
     directory: '/'
     schedule:
       interval: daily
-      time: 00:01
+      time: '00:01'
       timezone: America/Chicago
     groups:
       actions-deps:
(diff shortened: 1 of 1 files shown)


JTarasovic (approved):
_le sigh_

elliottneilclark:
https://noyaml.com/

PR #2132, opened 2025-06-13
feat(verify): async image pull for existing tests

JTarasovic:


elliottneilclark:
There are two different error modes that I can see. One is that a tuple with an empty string doesn't match. The other is that some array of atoms doesn't match.

```
(MatchError) no match of right hand side value: {"", 137}
```
```
(MatchError) no match of right hand side value: [:ollama]
```

elliottneilclark (approved):

PR #2139, opened 2025-06-14
fix: maybe_add_config was flakey because it re-used session

elliottneilclark:
Description:
This changes from re-using the a single session to configure the battery, to using the returned session from fill_in_name

Test Plan:
will run on CI with label: int-test


elliottneilclark:
Hmmmm this makes it better but might not solve this. Blah

JTarasovic:
Yeah. It needs to use fill_in_name. 

If you want, we can remove the config from forgejo for now and I can get it working next week.

PR #2166, opened 2025-06-18
feat: build optimized keycloak images

JTarasovic:
Added here in the batteries-included repo. I can move it to a separate repo if needed.


docker/keycloak.dockerfile
```diff
@@ -0,0 +1,21 @@
+ARG KC_VERSION=latest
```
elliottneilclark, line 1: Rather than latest can we take this in from the registry? So the kc version would be whatever the default tag is for the base keyckoak image.

elliottneilclark (approved):

docker/keycloak.dockerfile
JTarasovic: It's just a placeholder. The workflow passes the version as a `--build-arg`.

PR #2177, opened 2025-06-19
ci: automatically build on new upstream keycloak tags

JTarasovic:


bin/bi-registry
```diff
@@ -202,6 +203,78 @@ parse_params() {
...
+        [[ "${#tags[@]}" -eq 0 ]] && break
+
+        # otherwise add the local tags to the outer array
+        all_tags+=("${tags[@]}")
+        # and update the "last" parameter
+        last="${tags[-1]}"
+    done
+
+    echo "${all_tags[@]}"
+}
```
elliottneilclark, line 235: This functionality is already in the registry-tool. What do you think about using the registry list for this.

So we have

battery_included_keycloak & keycloak in the registry. Every night, the registry will be updated, which means that sometimes the registry will have a new tag for Keycloak that's not present for battery_included_keycloak. We can trigger a new workflow whenever the set of tags for the two registry keys is not the same.

bin/bi-registry
```diff
@@ -202,6 +203,78 @@ parse_params() {
...
+        # we don't need old versions so skip em
+        # we should periodically update this to limit the number of requests below
+        major=$(echo "$tag" | cut -d'.' -f1)
+        [[ "$major" -lt 24 ]] && continue
+
+        # we need the version if we don't already have it
+        curl --fail --silent -o /dev/null \
+            -H "Authorization: Bearer ${token}" \
+            -H "Accept: application/vnd.oci.image.index.v1+json" \
+            -I "https://ghcr.io/v2/${keycloak_repo}/manifests/${tag}" ||
```
elliottneilclark, line 269: Shouldn't we use the registry for tag listing/set operations?

PR #2193, opened 2025-06-20
test: try and make battery install worker most stable

elliottneilclark:
Description:
- Always re-use the session
- After installing or unisntalling a battery, trigger a deploy to cache the stat

Test Plan:
- CI


platform_umbrella/apps/verify/test/support/battery_install_worker.ex
```diff
@@ -53,8 +53,7 @@ defmodule Verify.BatteryInstallWorker do
       |> assert_has(Query.text("#{battery.name} Battery", minimum: 1))
       |> maybe_add_config(config)
       |> click(Query.text("Install Battery"))
-      # we have to pause a bit here for the install to actually take
-      |> sleep(1_000)
```
JTarasovic, line 57: I think we still need a small sleep here before triggering. If you click "Install" and navigate away very quickly, I've seen it not install.

JTarasovic (approved):

platform_umbrella/apps/verify/test/support/battery_install_worker.ex
elliottneilclark: The trigger_k8s_deploy includes a visit and a click and a sleep. Will that not be enough time?

JTarasovic: Sorry, I just refreshed. It's the navigating away that's the problem. I've watched it do that and not install the battery.

PR #2221, opened 2025-06-24
build: try and stop double trigger/cancel

elliottneilclark:
Description:
Lots of PR's are having issues with canceled jobs that never complete. I think that's because we are canceling too many which is happening because we are triggering too many. This removes the 'on' triggers and canceling logic to try

Test Plan:
- CI

### DIFF
--- a/.github/workflows/bix_docker.yml
+++ b/.github/workflows/bix_docker.yml
@@ -7,15 +7,13 @@ name: BIX Docker
       - master
     tags:
       - '*'
-
   pull_request:
     branches:
       - master
-    types:
-      - opened
-      - synchronize
-      - labeled
-      - reopened
+  # Allow manual dispatch for the times
+  # when we want to run int-tests after
+  # opening a PR
+  workflow_dispatch:
 
 permissions:
   id-token: write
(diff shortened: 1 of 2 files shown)


.github/workflows/bix_docker.yml
```diff
@@ -11,11 +11,6 @@ name: BIX Docker
   pull_request:
     branches:
       - master
-    types:
-      - opened
-      - synchronize
-      - labeled
-      - reopened
```
JTarasovic, line 18: opened, synchronized and reopened are the default types. Labeled is added so we're able to add the label. I guess it's not really any additional work to just manually re-run the workflow after we add the label to our own PRs. It would obviate the need for the concurrency groups below.

The other thing we could try is to remove the "default" labels from the dependabot.yml file. The documentation makes it sound like if you specify any labels there, it will not add the default labels but looking at the jobs that get kicked off, that doesn't seem accurate?

JTarasovic (approved):

.github/workflows/bix_docker.yml
elliottneilclark: > opened, synchronized and reopened are the default types. 

While they are the defaults there's something different about how they are types are de-duplicated when not listed. I'm reaching here.

>  just manually re-run the workflow after we add the label to our own PRs.

I added that, thanks for the call out.

PR #2230, opened 2025-06-24
fix(verify): try to robustify keycloak tests once more

JTarasovic:
- [x] creates an assertion for pod completion / success (useful for e.g. initdb)
- [x] allows using the workload statuses to determine if workload is "ready" as opposed to just "running"
- [x] use these to track each step of keycloak bootstrap


JTarasovic:
3 runs here with no sso / keycloak failures. Alternated between the two locally the entire time as well with no failures.

elliottneilclark (approved):
Thank you!

PR #2235, opened 2025-06-25
fix(keycloak): redeploy on config change

JTarasovic:


platform_umbrella/apps/common_core/lib/common_core/resources/sso/keycloak.ex
```diff
@@ -149,7 +157,8 @@ defmodule CommonCore.Resources.Keycloak do
                 },
                 %{"name" => "KEYCLOAK_ADMIN_PASSWORD", "valueFrom" => B.secret_key_ref("keycloak", "admin-password")},
                 %{"name" => "KC_DB_USERNAME", "valueFrom" => B.secret_key_ref(secret_name, "username")},
-                %{"name" => "KC_DB_PASSWORD", "valueFrom" => B.secret_key_ref(secret_name, "password")}
+                %{"name" => "KC_DB_PASSWORD", "valueFrom" => B.secret_key_ref(secret_name, "password")},
+                %{"name" => "CM_HASH", "value" => cm_hash}
```
elliottneilclark, line 161: This should already be handled by ResourceGenerator. Is it possible to make that do the work?

https://github.com/batteries-included/batteries-included/blob/0b6365bcb6376f925f0a585aabc8fc066a86e627/platform_umbrella/apps/common_core/lib/common_core/resources/resource_generator.ex#L106

Basically, for resources that mount or refer to secrets or config maps, we hash those adding the value as an annotation.

JTarasovic: Yeah, lemme check that.

elliottneilclark (approved):

PR #2246, opened 2025-06-26
feat(bi): allow running multiple kind clusters

JTarasovic:
Offsets the metallb ip address pools based on the existing containers so that multiple clusters can be run simultaneously.

### Test plan:
- [x] cluster with gateway is still accessible
- [x] dev cluster still works
- [x] int-test still works
- [x] combinations of the above are all accessible simultaneously

What other scenarios need testing?

### DIFF
--- a/bi/pkg/cluster/kind/net.go
+++ b/bi/pkg/cluster/kind/net.go
@@ -7,7 +7,7 @@ import (
 	"log/slog"
 	"net"
 
-	"github.com/dghubble/ipnets"
+	"github.com/apparentlymart/go-cidr/cidr"
 
 	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/api/types/network"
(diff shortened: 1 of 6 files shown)


bi/pkg/cluster/kind/net.go
```diff
@@ -7,7 +7,7 @@ import (
 	"log/slog"
 	"net"
 
-	"github.com/dghubble/ipnets"
+	"github.com/apparentlymart/go-cidr/cidr"
```
JTarasovic, line 10: We were already using this `cidr` package elsewhere so remove the `ipnets` module that's only used here as the functionality of `cidr` is a superset.

bi/pkg/cluster/kind/net_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func Test_split(t *testing.T) {
+	testCases := []struct {
+		sub, expected string
+		existing      int
+		errExpected   bool
+	}{
+		{sub: "172.0.0.0/8", expected: "172.0.1.0/24", existing: 0},
```
JTarasovic, line 17: These tests were added with the existing functionality and not changed (except for adding the offset) so we're getting the same results with `cidr` as with `ipnets`.

PR #2288, opened 2025-07-07
feat: add theme to keycloak build

elliottneilclark:
Summary:
- Add a bi-keycloak binary that will trigger missing, build images, and push
- Move `version_tag` and `docker_hash` to the common-functions.sh file
- Add keycloakify theme to the mono-repo
- Make it possible for our keycloak image tags to have -versionhash
- Add maven and java for keycloak build.... yeah
- Add the keycloak theme directory to the builds
- Add keycloak-theme to dependabot weekly
- Add the themes for most of the current login flow
- Add keycloak build to bix-docker

Test Plan:
Visual Through story book

image
image
image

### DIFF
--- a/bin/bi-registry
+++ b/bin/bi-registry
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 
 set -Eeuo pipefail
-set -o nounset # Error on unset variables
 
 # Get where the current script is located
 SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd -P)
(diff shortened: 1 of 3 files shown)


bin/bi-registry
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 set -Eeuo pipefail
+set -o nounset # Error on unset variables
```
JTarasovic, line 4: How is this different from the -u above?

JTarasovic (approved):

PR #2304, opened 2025-07-09
fix(verify): miscellaneous fixes for home-base e2e test

JTarasovic:
I'll merge in the AM.


platform_umbrella/apps/verify/test/support/kind_install_worker.ex
```diff
@@ -0,0 +1,202 @@
+defmodule Verify.KindInstallWorker do
```
JTarasovic, line 1: I moved this in a separate commit which should have made the diff easier but you can look at the individual commits.

platform_umbrella/apps/verify/lib/verify/application.ex
```diff
@@ -13,8 +13,7 @@ defmodule Verify.Application do
       Verify.SessionURLAgent,
       {Task.Supervisor, [name: Verify.TaskSupervisor]},
       {CommonCore.Installs.Generator, [name: Verify.Installs.Generator]},
-      {Verify.KindInstallWorker, [name: Verify.KindInstallWorker]},
-      # for BatteryInstall & ImagePull workers
+      # for workers
       {Registry, keys: :unique, name: Verify.Registry}
```
JTarasovic, line 17: We're starting a new worker per test module now. That'll help with state mgmt and cleanup especially as we'll have multiple clusters per module.

elliottneilclark (approved):

PR #2305, opened 2025-07-09
feat(verify): home-base e2e test

JTarasovic:
depends on: #2304


bi/pkg/specs/get_spec.go
```diff
@@ -24,9 +24,8 @@ func GetSpecFromURL(specURL string) (*InstallSpec, error) {
 		return readLocalFile(parsedURL)
 	case "https":
 		return readRemoteFile(parsedURL)
-		// Only download on HTTP for URLs that string contain 127.0.0.1
-	case "http":
-		if strings.Contains(parsedURL.Host, "127.0.0.1") || strings.Contains(parsedURL.Host, "127-0-0-1.batrsinc.co") {
+	case "http": // Only download on HTTP if home-base is running locally
+		if strings.Contains(parsedURL.Host, "127.0.0.1") || strings.Contains(parsedURL.Host, "batrsinc.co") {
```
elliottneilclark, line 28: This will allow http for all installs since they all have batrsinc urls. Do we need to accept other ip ranges as local?

JTarasovic: Wouldn't they be home.batteriesincl.com?

JTarasovic: I think I have a reasonable solution.

platform_umbrella/apps/home_base/priv/raw_files/common.sh
```diff
@@ -13,7 +13,6 @@ REPO="batteries-included/batteries-included"
 # Locations
 VERSION_LOC="${BASE_DIR}/share/bi/${BI}-${BI_VERSION_TAG}"
 INSTALL_DIR="${BASE_DIR}/bin"
-BI_LOC="${INSTALL_DIR}/${BI}"
```
JTarasovic, line 16: During the big refactor, this got changed to `VERSION_LOC`.

platform_umbrella/apps/verify/test/home_base_test.exs
```diff
@@ -53,16 +54,57 @@ defmodule Verify.HomeBaseTest do
...
+  defp navigate_to_home_base(session) do
+    session =
+      session
+      |> visit("/traditional_services")
+      |> assert_has(table_row(text: "home-base-"))
+
+    id = text(session, Query.css("table tr td:first-child"))
+    link = Query.link("running_service_#{id}")
+
+    {attr(session, link, "href"), session |> click(link) |> last_tab()}
```
JTarasovic, line 98: We need the home-base URL so we can pass the host as allowed when we call `bi start`

PR #2310, opened 2025-07-10
feat(bi): allow specifying insecure (HTTP) hosts

JTarasovic:


bi/pkg/specs/get_spec.go
```diff
@@ -35,6 +36,28 @@ func GetSpecFromURL(specURL string) (*InstallSpec, error) {
...
+
+	// allow download on HTTP if home-base is running locally
+	case strings.Contains(host, "127.0.0.1"):
+		return true
+
+	case strings.Contains(host, "127-0-0-1.batrsinc.co"):
+		return true
+
+	// or if user has allowed the specific host
+	case slices.ContainsFunc(allowedHosts, func(s string) bool {
```
JTarasovic, line 51: And here's the logic we need to allow the insecure home-base during integration testing.

PR #2368, opened 2025-07-25
feat(gateway_api): add initial resources

JTarasovic:


platform_umbrella/apps/common_core/lib/common_core/batteries/system_battery.ex
JTarasovic, line 1: These are just some updates for the most recent version of istio. It doesn't bootstrap without these if the gateway APIs are installed.

PR #2369, opened 2025-07-25
chore(istio): update crds

JTarasovic:


elliottneilclark (approved):
I'm so glad that we have istio so I can truly know that I have more than enough crds in my life.

PR #2377, opened 2025-07-28
fix: osx networking for container stacks

### LINKED ISSUE #2365: Get OSX networking finished for all the top container stacks
On Linux, Docker chooses an IP range for containers. That is appropriately routed on the machine via iptables. That doesn't happen on OSX. Mac does provide a VPN capability, and WireGuard is a great open-source client. We have already started a WireGuard container that we can use to route traffic. However, each of the different Mac container solutions has slightly different networking solutions. This issue is to resolve the remaining problems with the most popular OS X container networking stacks.

- [ ] Docker desktop
- [ ] Podman see: #757
- [ ] colima
- [ ] Apple container?

## When It Works

`bix local start bootstrap/local.spec.json` will start then show an HTTP server and a set of commands to run to get and start the wireguard vpn.

krushnarout:
```
export DOCKER_HOST=unix:///var/folders/0r/kjdxt_fs6flckdfrf985qc940000gn/T/podman/podman-machine-default-api.sock && ./bin/bi start bootstrap/local.spec.json -v debug  
```

VPN connection

```
./bin/bi vpn config -o wg0-local.conf local

sudo wg-quick up ./wg0-local.conf

curl -s -o /dev/null -w "status: %{http_code}\n" https://www.batteriesincl.com/

```

Test demo:

https://github.com/user-attachments/assets/2f3eb03f-b6e4-46d0-a15c-1280833cde51

closes #2365
/claim #2365

### DIFF
--- a/bi/pkg/cluster/kind/control_server.go
+++ b/bi/pkg/cluster/kind/control_server.go
@@ -1,0 +1,107 @@
+package kind
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"net/http"
+	"strings"
+	"time"
+)
+
+// ControlServerChecker verifies that the control server is accessible
+type ControlServerChecker struct {
+	baseURL string
+}
+
+func NewControlServerChecker(baseURL string) *ControlServerChecker {
+	return &ControlServerChecker{
+		baseURL: baseURL,
+	}
+}
+
+func (csc *ControlServerChecker) WaitForControlServer(ctx context.Context, timeout time.Duration) error {
+	if strings.Contains(csc.baseURL, "localhost") {
+		slog.Info("Skipping control server check for localhost - VPN connection required")
+		return fmt.Errorf("localhost check skipped - VPN connection required")
+	}
+	
+	slog.Info("Waiting for control server to become accessible", 
+		slog.String("url", csc.baseURL),
+		slog.Duration("timeout", timeout))
+	
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+	
+	ticker := time.NewTicker(5 * time.Second)
+	defer ticker.Stop()
+	
+	client := &http.Client{
+		Timeout: 10 * time.Second,
+	}
+	
+	for {
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("timeout waiting for control server to become accessible")
+		case <-ticker.C:
+			if err := csc.checkControlServer(client); err != nil {
+				slog.Debug("Control server not yet accessible", slog.String("error", err.Error()))
+				continue
+			}
+			
+			slog.Info("Control server is now accessible", slog.String("url", csc.baseURL))
+			return nil
+		}
+	}
+}
+
+func (csc *ControlServerChecker) checkControlServer(client *http.Client) error {
+	if strings.Contains(csc.baseURL, "localhost") {
+		return fmt.Errorf("localhost check skipped - VPN connection required")
+	}
+	
+	resp, err := client.Get(csc.baseURL + "/health")
+	if err != nil {
+		return fmt.Errorf("failed to connect to control server: %w", err)
+	}
+	defer resp.Body.Close()
+	
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("control server returned status %d", resp.StatusCode)
+	}
+	
+	return nil
+}
+
+// GetControlServerURL returns the URL where the control server should be accessible
+func GetControlServerURL(runtimeInfo *ContainerRuntimeInfo) string {
+	switch runtimeInfo.Runtime {
+	case RuntimePodman:
+		if runtimeInfo.HostIP != "localhost" {
+			return fmt.Sprintf("http://%s", runtimeInfo.HostIP)
+		}
+		return "http://localhost"
+	case RuntimeColima:
+		if runtimeInfo.HostIP != "localhost" {
+			return fmt.Sprintf("http://%s", runtimeInfo.HostIP)
+		}
+		return "http://localhost"
+	case RuntimeDockerDesktop:
+		return "http://localhost"
+	default:
+		return "http://localhost"
+	}
+}
+
+func PrintWhenItWorksInstructions(runtimeInfo *ContainerRuntimeInfo, clusterName string) {
+	fmt.Println("\nSUCCESS! Your Batteries Included cluster is ready!")
+	fmt.Println(strings.Repeat("=", 60))
+	
+	controlServerURL := GetControlServerURL(runtimeInfo)
+	
+	fmt.Printf("\nControl server: %s\n", controlServerURL)
+	fmt.Printf("Connect via VPN: bi vpn config -o wg0-%s.conf %s\n", clusterName, clusterName)
+	fmt.Printf("For help: bi --help\n")
+	fmt.Println(strings.Repeat("=", 60))
+}
--- a/bi/pkg/cluster/kind/net.go
+++ b/bi/pkg/cluster/kind/net.go
@@ -44,17 +44,50 @@ func GetMetalLBIPs(ctx context.Context) (string, error) {
 // Return the bottom half of the network, such as 172.18.1.0/24
 func split(ipNet *net.IPNet, count int) (*net.IPNet, error) {
 	shift := calculateShift(ipNet)
-	return cidr.Subnet(ipNet, shift, count+1)
+	
+	// Calculate maximum number of subnets we can create
+	maxSubnets := 1 << shift
+	
+	// Use count + 1 as the subnet number to avoid subnet 0
+	// This ensures we get sequential subnets starting from 1
+	subnetNum := count + 1
+	
+	// Check if we can fit the requested subnet
+	if subnetNum >= maxSubnets {
+		return nil, fmt.Errorf("cannot allocate subnet %d, maximum subnets: %d", subnetNum, maxSubnets)
+	}
+	
+	subnet, err := cidr.Subnet(ipNet, shift, subnetNum)
+	if err != nil {
+		return nil, fmt.Errorf("failed to allocate subnet %d: %w", subnetNum, err)
+	}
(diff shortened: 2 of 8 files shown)


elliottneilclark:
I tried it on my machine, and I was un-able to get control-server to start.

```

SUCCESS! Your Batteries Included cluster is ready!
============================================================

Control server: http://localhost
Connect via VPN: bi vpn config -o wg0-local.conf local
For help: bi --help
============================================================
13:32:55.576 INF kind/kind.go:198 Skipping localhost control server check - VPN connection required for access
13:32:55.579 DBG kind/net.go:24 Found kind networks:  networks=2 existing=2
Usage:
  bi start [install-slug|install-spec-url|install-spec-file] [flags]

Flags:
  -h, --help             help for start
      --skip-bootstrap   Skip bootstrapping the cluster

Global Flags:
  -c, --color              Use color in logs (default true)
      --config string      config file (default is $XDG_CONFIG_HOME/bi/bi.yaml)
  -v, --verbosity string   Log level (debug, info, warn, error (default "WARN")

13:32:55.579 ERR cmd/root.go:39 Error executing command error="unable to start kube provider: error adding metal ips: error getting metal lb ips: error splitting network: cannot allocate subnet 3, maximum subnets: 2"
```

Podman version:
https://gist.github.com/elliottneilclark/c67141b2809a9b086f859eec31b3df67

Full Log:
https://gist.github.com/elliottneilclark/b8b7a6d09cfd9096be16635df7752ce1

Podman Networks:
https://gist.github.com/elliottneilclark/1675f673b9befa6d223cd4dc6d18d68a

elliottneilclark:
I also noticed in your demo that you're curling battierincl.com. Instead, you should show that you can access the control server that's running and has a MetalLB IP.

For example here's me starting on linux: https://gist.github.com/elliottneilclark/dc2c9ba789536300006cd06b9c196876

That gives a kind network like this: https://gist.github.com/elliottneilclark/6483b2063d088c1e37bfa712cd5bb2e6

So we get:

```
Welcome to your Batteries Included platform: http://control.172-18-2-1.batrsinc.co
```

Then we can test that the control server is up.

```
curl -vvv http://control.172-18-2-1.batrsinc.co/healthz
```

elliottneilclark:
I checked out a2940e85755011afc75f8332f1ab1392bddbce17 and all the pods are running.
https://gist.github.com/elliottneilclark/f7d7ac646cf56174f22492465f06563b

However the http check to control server never worked.
https://gist.github.com/elliottneilclark/98ac95974188026bc57ffd9aab2fc047

krushnarout:
@elliottneilclark What I found is that when it successfully connects to the cluster.

```
11:41:56.618 INF kube/kube.go:160 Successfully connected to cluster
```

After that, you need to connect to the VPN, because the control server is only accessible through the VPN

```
./bin/bi vpn config -o wg0-local.conf local
sudo wg-quick up ./wg0-local.conf
```

After some time, It will work!

```
Welcome to your Batteries Included platform: http://control.172-18-1-1.batrsinc.co
```

elliottneilclark:
I've tried it a couple of times and I can never get it to start.

For me the cluster starts, the control server starts, nothing can connect to the control server, on of the revisions of the controlserver deployment is stuck trying to update the database on init. Even if I connect to the wireguard on the mac, curl can't connect to the controlserver.

krushnarout:
> I've tried it a couple of times and I can never get it to start.
> 
> For me the cluster starts, the control server starts, nothing can connect to the control server, on of the revisions of the controlserver deployment is stuck trying to update the database on init. Even if I connect to the wireguard on the mac, curl can't connect to the controlserver.

@elliottneilclark I think there is something wrong on your side. Please try again with these commands:

```
# Stop any existing VPN connections
sudo wg-quick down wg0-local 2>/dev/null

# Clean up Kind clusters
kind delete cluster --name local

# Remove old VPN config files
rm -f wg0-local.conf
```

Build the updated binary

```
cd bi && go build -o ../bin/bi .
cd ..
```

Start the cluster 

```
TRACE=1 bin/bix local start bootstrap/local.spec.json
```

Check pods are running

```
kubectl --context kind-local get pods -A
```

Set up VPN connection

```
# Generate VPN config
./bin/bi vpn config -o wg0-local.conf local

# Connect to VPN
sudo wg-quick up ./wg0-local.conf
```

elliottneilclark:
Everything seems fine on my end.

```
podman machine stop
podman machine rm
podman machine init --cpus 6 --disk-size 300 -m 8000 --rootful
podman machine start
TRACE=1 bix local start bootstrap/local.spec.json 2>&1 | tee ~/test.log
```

Getting the networks, I think I see a clue:

```
podman network ls --format json
[
  {
    "name": "kind",
    "id": "e6a31ada4aa95c877ff1c07d64df67cabbe20e6b40df32a0b0f6c405381cbeaf",
    "driver": "bridge",
    "network_interface": "podman1",
    "created": "2025-07-30T16:05:06.217687788-05:00",
    "subnets": [
      {
        "subnet": "fc00:f853:ccd:e793::/64",
        "gateway": "fc00:f853:ccd:e793::1"
      },
      {
        "subnet": "10.89.0.0/24",
        "gateway": "10.89.0.1"
      }
    ],
    "ipv6_enabled": true,
    "internal": false,
    "dns_enabled": true,
    "ipam_options": {
      "driver": "host-local"
    }
  },
  {
    "name": "podman",
    "id": "2f259bab93aaaaa2542ba43ef33eb990d0999ee1b9924b557b7be53c0b7a1bb9",
    "driver": "bridge",
    "network_interface": "podman0",
    "created": "2025-07-30T16:15:01.80418704-05:00",
    "subnets": [
      {
        "subnet": "10.88.0.0/16",
        "gateway": "10.88.0.1"
      }
    ],
    "ipv6_enabled": false,
    "internal": false,
    "dns_enabled": false,
    "ipam_options": {
      "driver": "host-local"
    }
  }
]
```

Notice that the kind network has the following definition:

```
      {
        "subnet": "10.89.0.0/24",
        "gateway": "10.89.0.1"
      }
```


So the `10.89.0.1` is the network gateway. However that's also the ip that's being given to the ingress load-balancer.

```
kubectl get svc --all-namespaces -w
NAMESPACE       NAME                      TYPE           CLUSTER-IP     EXTERNAL-IP   PORT(S)                                      AGE
battery-base    metallb-webhook-service   ClusterIP      10.96.174.29   <none>        443/TCP                                      12m
battery-base    pg-controlserver-r        ClusterIP      10.96.86.103   <none>        5432/TCP                                     11m
battery-base    pg-controlserver-ro       ClusterIP      10.96.69.53    <none>        5432/TCP                                     11m
battery-base    pg-controlserver-rw       ClusterIP      10.96.18.225   <none>        5432/TCP                                     11m
battery-core    battery-control-server    ClusterIP      10.96.28.81    <none>        4001/TCP                                     12m
battery-core    cnpg-webhook-service      ClusterIP      10.96.189.95   <none>        443/TCP                                      12m
battery-istio   istio-ingressgateway      LoadBalancer   10.96.228.43   10.89.0.1     15021:32333/TCP,80:31672/TCP,443:31093/TCP   12m
battery-istio   istiod                    ClusterIP      10.96.199.93   <none>        15010/TCP,15012/TCP,443/TCP,15014/TCP        12m
default         kubernetes                ClusterIP      10.96.0.1      <none>        443/TCP                                      12m
kube-system     kube-dns                  ClusterIP      10.96.0.10     <none>        53/UDP,53/TCP,9153/TCP                       12m
```

MetalLB should never be able to give out an ip that is in use for the network gateway.

bi/pkg/cluster/kind/control_server.go
```diff
@@ -0,0 +1,107 @@
...
+	
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+	
+	ticker := time.NewTicker(5 * time.Second)
+	defer ticker.Stop()
+	
+	client := &http.Client{
+		Timeout: 10 * time.Second,
+	}
```
elliottneilclark, line 41: When setting up the HTTP client, we'll need to add the WireGuard transport here. 

See an example here: https://github.com/noisysockets/noisysockets/blob/9912d4c4a67b8b539f8bfa3f81b2a5dc72c8d89c/examples/router/main.go#L84

or here: https://github.com/batteries-included/batteries-included/blob/58a8e70f899e9df03d47df029dba546054193e8f/bi/pkg/kube/kube.go#L81-L103

bi/pkg/cluster/kind/net.go
```diff
@@ -41,23 +43,164 @@ func GetMetalLBIPs(ctx context.Context) (string, error) {
...
+// Return an available subnet for MetalLB, such as 172.18.1.0/24
 func split(ipNet *net.IPNet, count int) (*net.IPNet, error) {
+	ones, bits := ipNet.Mask.Size()
+	slog.Debug("Split network details", 
+		slog.String("baseNetwork", ipNet.String()),
+		slog.Int("maskOnes", ones),
+		slog.Int("maskBits", bits))
+	
 	shift := calculateShift(ipNet)
-	return cidr.Subnet(ipNet, shift, count+1)
```
elliottneilclark, line 47: Notice how we never gave out the bottom half. That's because almost all network layouts need a gateway and they usually have .1 reserved for that.

elliottneilclark (changes requested):
I think giving out the gateway address is likely the issue. However we'll also need the HTTP client to work reliably without the user needing to connect to the vpn.

elliottneilclark:
Any progress that you need me to test?

krushnarout:
> Any progress that you need me to test?

I need some time, I'm currently fixing and testing it.

krushnarout:
@elliottneilclark I was having trouble after https://github.com/batteries-included/batteries-included/pull/2377#discussion_r2243996764 and couldn't get it working without a vpn, I see the pr https://github.com/batteries-included/batteries-included/pull/2420 found out what's the problem I was making. Nice too see this working 🎉

Thank you for the tip :)

PR #2429, opened 2025-08-06
build: use locally built elixir

elliottneilclark:
ElixirLS and others can do go to defintion all the way into elixir standard library if the paths are valid. So installing from source makes that work. This updates to use a reference to the git tag, so that asdf installs from source.

This also bumps the otp version


JTarasovic (approved):

elliottneilclark:
This wasn't worth it. asdf doesn't seem to make promises about ordering of install for plugins. So getting erlang to install before elixir is tricky.

PR #2430, opened 2025-08-06
fix(istio): only spin up waypoints if needed

JTarasovic:


platform_umbrella/apps/common_core/lib/common_core/resources/istio/gateways.ex
```diff
@@ -37,5 +38,6 @@ defmodule CommonCore.Resources.Istio.Gateways do
       "gatewayClassName" => "istio-waypoint",
       "listeners" => [%{"name" => "mesh", "port" => 15_008, "protocol" => "HBONE"}]
     })
+    |> F.require_battery(state, :sso)
```
elliottneilclark, line 41: I believe we want `:keycloak` here, since users can install keycloak and select a realm for a traditional service or notebook.

elliottneilclark (changes requested):
Do we need the namespace labels to be conditional too?

elliottneilclark (approved):

PR #2440, opened 2025-08-08
feat: add gpt-oss to the ollama models

elliottneilclark:
Summary:
- Add both variants of gpt oss
- Move the cpu requirements for sizes down.

Test:
- model instances


elliottneilclark:
My laptop doesn't have enough memory to load them so I'll wait until I get home to land.

PR #2461, opened 2025-08-13
chore(cnpg): add initial barman plugin resources (and test)

JTarasovic:
This just creates a battery that will spin up the barman plugin.
The migration from the legacy backup config to barman will be in a followup.

### DIFF
--- a/platform_umbrella/apps/common_core/lib/common_core/batteries/catalog.ex
+++ b/platform_umbrella/apps/common_core/lib/common_core/batteries/catalog.ex
@@ -83,6 +83,14 @@ defmodule CommonCore.Batteries.Catalog do
     },
     %CatalogBattery{
       group: :data,
+      type: :cloudnative_pg_barman,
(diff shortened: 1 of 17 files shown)


platform_umbrella/apps/common_core/lib/common_core/batteries/catalog.ex
```diff
@@ -81,6 +81,14 @@ defmodule CommonCore.Batteries.Catalog do
       description:
         "Redis is an in-memory data structure store, used as a database, cache, message broker, and streaming engine."
     },
+    %CatalogBattery{
+      group: :data,
+      type: :cloudnative_pg_barman,
+      dependencies: [:battery_ca],
+      name: "Barman Cloud Plugin for CNPG",
+      description: "Backup and Recovery Manager for PostgreSQL",
+      allowed_usages: ~w(internal_dev internal_int_test)a
```
JTarasovic, line 90: Marked as internal for now.

PR #2462, opened 2025-08-13
chore: sort failed paths to the top

JTarasovic:


JTarasovic:
Sorts the failed resources to the top for easier debugging / troubleshooting.

image

PR #2488, opened 2025-08-19
docs: AI documentation dump

elliottneilclark:
Summary:
- Add Local AI docs
- Add AI Manifesto
- Change BI so that it does in container validation making the doc much easier.

Test Plan:
- Editing
- Netlify preview

PR #2516, opened 2025-08-22
chore: prepare for 1.4.0 development

elliottneilclark:
Summary:
- Release the new cli so that docs are correct
- bix s set-version 1.4.0

Test Plan:
- CI


elliottneilclark:
This will go in after #2488 since that includes the cli change to make the docs correct.

JTarasovic (approved):

elliottneilclark:
Yeah that bi change wasn't supposed to be here. Fixed.

PR #2529, opened 2025-08-26
refactor(build): pull versions from existing registries

JTarasovic:
We keep track of versions and images in a few place but are consolidating to single repos/registries. This continues that work by moving base image version definitions to the registries.


elliottneilclark (approved):
Single source of versions, Yay!

PR #2533, opened 2025-08-27
feat(cnpg): create custom postgres images

JTarasovic:


docker/postgres.dockerfile
```diff
@@ -0,0 +1,34 @@
...
+        --no-install-recommends \
+        -o Dpkg::::="--force-confdef" \
+        -o Dpkg::::="--force-confold" \
+        postgresql-common && \
+    sed -ri 's/#(create_main_cluster) .*$/\1 = false/' /etc/postgresql-common/createcluster.conf && \
+    apt-get install -y \
+        --no-install-recommends \
+        -o Dpkg::::="--force-confdef" \
+        -o Dpkg::::="--force-confold" \
+        "postgresql-${PG_MAJOR}=${PG_VERSION}*" && \
```
elliottneilclark, line 29: While we are here, we should install:

- pgaudit
- pgvector
- pg-failover-slots
- pg_stat_monitor
- postgis-3

They seem to be in the form of `postgres-${PG_MAJOR}-extension-name`

elliottneilclark (changes requested):
Failover slots at least is needed in our prod. I think stat monitor is as well.

docker/postgres.dockerfile
JTarasovic: pg_stat_monitor isn't available in the standard packages and isn't installed in the cnpg images. We can add that later but I'd like to get it working initially and land it before jumping into additional features. Everything else is added now.

elliottneilclark: Totally, I was just going by this: https://www.postgresql.org/download/products/6-postgresql-extensions/

elliottneilclark (approved):

PR #2563, opened 2025-09-03
ci: allow registry update job to read BI packages

JTarasovic:


bin/lib/common-functions.sh
```diff
@@ -21,6 +21,8 @@ safe_colors() {
 }
 
 setup_trace() {
+    # set TRACE if GHA debug logging is enabled
+    [[ ${RUNNER_DEBUG:-0} -eq 1 ]] && TRACE=1
```
JTarasovic, line 25: This will allow us to enable debug logging when we re-run a job.

Image

PR #2576, opened 2025-09-04
feat: add RoboSRE

elliottneilclark:
This is the PR to implement all of RoboSRE. It will end up being a stacked diff pr where each commit builds on top of each other to build the whole feature since it's not useful until there's a lot of code that works.

- [x] Add the data model
- [x] Add the OTP GenServer's needed
- [x] Wire up stale resource detection as a part of robosre.
- [x] Add UI for issues list
- [x] Add UI for a single issue
- [x] Wire up stuck kube state via drift detection of whats in kube state vs what the client gets.


elliottneilclark:
image

PR #2623, opened 2025-09-15
feat!: upgrade to ferretdb 2.0+

JTarasovic:
This works end-to-end with a dump from 1 and restore to 2.5.0.


elliottneilclark (approved):
Now that v2 does users have you given any thought on how we expose that and possibly give a mongo uri/useename/password secret for the users

JTarasovic:
Ferret just passes it along to postgres so whatever PG users/roles get created as ferret/mongo URIs.

PR #2630, opened 2025-09-16
ci: try to fix integration tests

JTarasovic:


elliottneilclark (approved):
This seems correct too.

PR #2635, opened 2025-09-17
ci: fix up failure logging / output

JTarasovic:
I put in a raise on battery install and got rage output and screenshot:

```
/tmp/bi-int-test
└── Elixir.Verify.ForgejoTest
    ├── 1758140504_elixir-verify-forgejotest.json
    └── battery-install-worker-failure-forgejo.png
```

### DIFF
--- a/.github/workflows/int_test.yml
+++ b/.github/workflows/int_test.yml
@@ -104,7 +104,9 @@ jobs:
         if: failure()
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
         with:
-          name: int-test-${{ github.run_id }}-${{ github.run_attempt }}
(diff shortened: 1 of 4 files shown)


.github/workflows/int_test.yml
```diff
@@ -104,7 +104,9 @@ jobs:
         if: failure()
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
         with:
-          name: int-test-${{ github.run_id }}-${{ github.run_attempt }}
+          name:
+            int-test-${{ matrix.partition }}-${{ github.run_id }}-${{
```
JTarasovic, line 108: This should allow multiple fails per run without shitting the bed.

PR #2667, opened 2025-09-22
fix: prevent spurious cs health check failures

JTarasovic:


elliottneilclark (approved):
~~Do we know why it's taking so long to get started?~~

Nevermind. Thanks for the info.

PR #2684, opened 2025-09-25
feat: add CommonCore.Metrics

elliottneilclark:
Summary:

- Add CommonCore.Metrics.Reporter that will report telemetry_metrics
  into CommonCore.Metrics.Store.
- Add CommonCore.Metrics.Store to get data and then return it formatted.
- Hook that up for ControlServer, ControlServerWeb, and HomeBaseWeb
- Hook up getting the metrics json from `bi rage`

Tests:
- Tests added
- int-tests added


JTarasovic (approved):
Would it behoove us to a library like [this](https://hexdocs.pm/prometheus_ex/Prometheus.html) or am I missing something?

PR #2724, opened 2025-10-01
fix: don't revert route hostnames on startup

JTarasovic:


platform_umbrella/apps/common_core/lib/common_core/state_summary/hosts.ex
```diff
@@ -189,11 +207,19 @@ defmodule CommonCore.StateSummary.Hosts do
     summary.kube_state
     |> Map.get(:service, [])
     |> ingress_ips_from_services(istio_namespace, ingress_name)
+    |> maybe_default_ip()
     |> Enum.sort(:asc)
   end
 
+  defp maybe_default_ip(ips) when is_nil(ips) or ips == [] do
```
JTarasovic, line 214: This is a cleaned up version of the logging from the test PR.

PR #2727, opened 2025-10-02
WIP: no review needed

JTarasovic:


bi/pkg/cluster/kind/kind.go
```diff
@@ -178,6 +179,22 @@ func (c *KindClusterProvider) Create(ctx context.Context, progressReporter *util
 
 // createClusterConfig creates a kind cluster configuration, with GPU support if available
 func (c *KindClusterProvider) createClusterConfig() *v1alpha4.Cluster {
+	patch, err := yaml.Marshal(map[string]interface{}{
```
elliottneilclark, line 182: Could we pipe this in with our log level from flags? So we add this whenever start or start-local get a --log=debug

JTarasovic: Yeah. It might be a little messy but shouldn't be too bad.

PR #2728, opened 2025-10-02
fix: log default IPs as error for troublshooting

JTarasovic:


elliottneilclark (approved):
Are we keeping enough logs to help debug in rage?

PR #2729, opened 2025-10-02
feat(bi): embed static assets

JTarasovic:


elliottneilclark (approved):
Nice catch. I haven't internalized all of the go std yet. I appreciate it

PR #2732, opened 2025-10-03
fix(trivy): updates including new registries

JTarasovic:


image_registry.yaml
```diff
@@ -31,19 +31,25 @@ aqua_trivy:
     - 0.62.1
   tag_regex: ^\d+\.\d+\.\d+$
 aqua_trivy_checks:
-  name: ghcr.io/aquasecurity/trivy-checks
+  name: mirror.gcr.io/aquasec/trivy-checks
```
elliottneilclark, line 34: Why change to `mirror.gcr.io`? My understanding is that google has deprecated it.

The helm has docker.io: https://github.com/aquasecurity/trivy/blob/main/helm/trivy/values.yaml#L5

The github ghcr.io/aquasecurity/trivy-db exists and is updated. 

If possible the ghcr service will usually have a highest rate limits for us.

PR #2743, opened 2025-10-06
chore(victoria_metrics): update images to try to avoid rate limits

JTarasovic:


JTarasovic:
I ran the VM int-test in a loop for a couple of hours on Friday evening without getting shut down.

elliottneilclark (approved):

